PR #1, opened 2019-06-11
Setup monorepo

YuCJ:


makefile
```diff
@@ -0,0 +1,26 @@
...
+SUBDIRS := $(wildcard packages/*/.)
+
+yarn-install:
+	yarn install
+
+# Run make dev in each subfoler
+# Should use flag `-j` to start multiple watch tasks in parallel and flag `-i` to prevent exiting process on error
+dev: yarn-install $(SUBDIRS)
+$(SUBDIRS):
+	$(MAKE) -C $@ $(MAKECMDGOALS)
```
nickhsine: Question:
Since we use `babel --watch` in each dir of `$(SUBDIRS)`, `make dev` would stop and listen to file changes at the first `$(SUBDIRS)`, such as `@twreporter/core`.
Hence, if we modify `packages/react-components/src/index.js`, babel won't transpile the file.

nickhsine:
LGTM 🥇

makefile
YuCJ: We need to use `--jobs` or `-j` flag when executing the root makefile to run its recipes in parallel. 
Ref. https://www.gnu.org/software/make/manual/html_node/Parallel.html

I think we can add a recipe like `dev-all`:

```makefile
dev-all:
  make dev -j
```

nickhsine: My thought is we could specify the package we want to `watch`.

Sample code:
```
dev-all: yarn-install $(SUBDIRS)

$(SUBDIRS):
  $(MAKE) -C $@ dev

dev:
ifneq (,$(wildcard packages/$(PKG)/))
  $(MAKE) -C packages/$(PKG)/. dev
else ifneq (, $(PKG))
  @echo $(PKG) is not existed
  @echo $(shell ls packages) are the package list you can use
else
  make -j dev-all
endif
```

We can run

```
// dev on one pkg
make PKG=core dev
make PKG=react-components dev

// dev on all pkgs
make dev
```

PR #11, opened 2019-07-22
Donation link with utm query string

nickhsine:
### Change Reason
https://kanbanflow.com/t/mkh8MzGM

### DIFF
--- a/packages/react-components/src/donation-link-with-utm.js
+++ b/packages/react-components/src/donation-link-with-utm.js
@@ -1,6 +1,8 @@
-import React from 'react'
-import externalLinks from '@twreporter/core/lib/constants/external-links'
+/* eslint node/no-deprecated-api: "warn" */
 import PropTypes from 'prop-types'
+import React from 'react'
+import url from 'url'
+import externalLinks from '@twreporter/core/lib/constants/external-links'
 
 export default function DonationLinkWithUtm(props) {
   const { children = null, utmMedium = '' } = props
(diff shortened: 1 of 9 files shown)


packages/react-components/src/donation-link-with-utm.js
```diff
@@ -0,0 +1,42 @@
...
+  const handleClick = e => {
+    e.preventDefault()
+
+    let donationURL = externalLinks.donation
+
+    try {
+      const utmSource = window.location.host
+      const utmCampaign = encodeURIComponent(window.location.pathname)
+      const search = `utm_source=${utmSource}&utm_medium=${utmMedium}&utm_campaign=${utmCampaign}`
+      const url = new URL(donationURL)
```
YuCJ: I found that IE and some version of edge do not support the URL object (or not support some properties)
https://developer.mozilla.org/en-US/docs/Web/API/URL

nickhsine: I will change to `import URL = require('url')`.

nickhsine: Updated. 
Please have a see the [last commit](https://github.com/twreporter/twreporter-npm-packages/pull/11/commits/e37da37ec867de1b23f30060b95539ad363a0426)

nickhsine:
Updated. 
Please have a see.

packages/react-components/src/donation-link-with-utm.js
YuCJ: But the code is run on browser? Or can we use library like `url-parse` or `qs`?

YuCJ: After discussion, the client will take the webpack bundled js file that include the `url` module given by Node. So it's okay to use node modules.

nickhsine: `url` would be bundled by `webpack`. 
Hence, the codes could be executed on browser(tested).

YuCJ:
LGTM 👍

nickhsine:
rebase commits.

PR #12, opened 2019-07-24
Fix bookmark-widget ssr error

YuCJ:
After this update, the component will take `slug` and `host` from window, and won't use them when server-side rendering.

### DIFF
--- a/packages/react-components/src/bookmark-widget/index.js
+++ b/packages/react-components/src/bookmark-widget/index.js
@@ -118,13 +118,13 @@ class BookmarkWidget extends React.PureComponent {
     /* TODO: Implement `status` for bookmark widget in redux reducer and action:
       There should be different states below for the bookmark widget status of an article:
         unknown: It has not checked the bookmark status yet
-        isChecking: The request to check was make but has not gotten the response yet
+        isChecking: The request to check was made but has not gotten the response yet
         bookmarked: The article is bookmarked
         notBookmarked: The article is not bookmarked
         invalid: There's an error or there's no valid authentication info
       The current code does not distinguish the `unknown`, `notBookmarked`, `isChecking`, and `invalid` situation.
       The best result is that we only send request to check bookmark when the status is `unknown`.
-      But now we send request when there's no bookmark data for this component in the redux store when `componentDidMount`, no matter the reason of it.
+      But now we send request when there's no bookmark data for this component in the redux store when `componentDidMount`, no matter what's the reason of it.
      */
     const articleSlug = _.get(this.props, 'articleMeta.slug')
     if (articleSlug && typeof articleSlug === 'string') {
--- a/packages/react-components/src/bookmark-widget/index.js
+++ b/packages/react-components/src/bookmark-widget/index.js
@@ -98,7 +98,10 @@ function getHostFromWindowLocation() {
       return defaultHost
     }
   } catch (err) {
-    console.warn('Error on getting `host` from `window.location`:', err)
+    console.warn(
+      'Error on getting `host` and `protocol` from `window.location`:',
+      err
+    )
   }
 }
 
@@ -109,29 +112,45 @@ class BookmarkWidget extends React.PureComponent {
     this.removeCurrentPageFromBookmarks = this.removeCurrentPageFromBookmarks.bind(
       this
     )
-    this.requestedBookmarkSlug = ''
+    this.requestedBookmarkSlugs = []
   }
 
-  componentWillMount() {
+  componentDidMount() {
     this.tryToFetchTheBookmarkOnce()
   }
 
   componentDidUpdate() {
     this.tryToFetchTheBookmarkOnce()
   }
+
+  componentWillUnmount() {
+    this.requestedBookmarkSlugs = []
+  }
+
   tryToFetchTheBookmarkOnce() {
     const articleSlug = _.get(this.props, 'articleMeta.slug')
     /* Do nothing if the article slug has not been fetched yet, or the bookmark of it has been requested */
-    if (articleSlug && articleSlug !== this.requestedBookmarkSlug) {
+    /* TODO: Implement `status` for bookmark widget in redux reducer and action:
+        There should be different states below for the bookmark widget status of an article:
+          unknown: It has not checked the bookmark status yet
+          isChecking: The request to check was make but has not gotten the response yet
+          bookmarked: The article is bookmarked
+          notBookmarked: The article is not bookmarked
+          invalid: There's an error or there's no valid authentication info
+        The current code is a workaround to identify the `unknown` situation (by logging we've sent request for an article or not).
+        We only want to send request to check bookmark when the status is `unknown`.
+     */
+    if (
+      articleSlug &&
+      !this.requestedBookmarkSlugs.some(value => value === articleSlug)
+    ) {
       const hostFromWindow = getHostFromWindowLocation()
-      if (this.checkIfThisArticleBookmarked()) {
-        /* Do not need to fetch bookmark if it has been already fetched (by SSR) */
-        this.requestedBookmarkSlug = articleSlug
-      } else {
+      if (!this.checkIfThisArticleBookmarked()) {
         const { jwt, userID, getSingleBookmark } = this.props
-        this.requestedBookmarkSlug = articleSlug
         getSingleBookmark(jwt, userID, articleSlug, hostFromWindow)
       }
+      /* Do not need to fetch bookmark if it has been already fetched (by SSR) */
+      this.requestedBookmarkSlugs.unshift(articleSlug)
     }
   }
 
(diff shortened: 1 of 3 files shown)


YuCJ:
Update according to review. The logic of this component still seems not so elegant. I think the reducer of bookmarks and bookmark widget need to be refactor to solve the problem. And it can prevent duplicate trying to check an article is bookmark or not.

packages/react-components/src/bookmark-widget/index.js
```diff
@@ -80,50 +80,58 @@ const BookmarkImgDesktop = styled.div`
...
-  getCurrentSlug() {
-    const slugFromProps = _.get(this.props, 'articleMeta.slug')
-    if (slugFromProps) return slugFromProps
-    if (typeof window !== 'undefined') {
-      const pathname = _.get(window, 'location.pathname', '')
-      return _.get(pathname.match(/(?:\w+-)*\w+$/), [0]) || '' // take 'xxx-xxx-xx' from string '/oo-xx/oxox/xxx-xxx-xx'
+  tryToFetchTheBookmarkOnce() {
+    const articleSlug = _.get(this.props, 'articleMeta.slug')
+    /* Do nothing if the article slug has not been fetched yet, or the bookmark of it has been requested */
+    if (articleSlug && articleSlug !== this.requestedBookmarkSlug) {
```
nickhsine: why do we need `articleSlug !== requestedBookmarkSlug` condition here?

packages/react-components/src/bookmark-widget/index.js
```diff
@@ -211,27 +227,20 @@ class BookmarkWidget extends React.PureComponent {
 }
 
 BookmarkWidget.defaultProps = {
-  articleMeta: {},
+  articleMeta: null,
```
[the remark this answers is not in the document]
YuCJ: I will take off the `isRequired`. Because the `articleMeta` may be asynchronously given and might not be fetched successfully.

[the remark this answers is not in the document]
nickhsine: If we remove `isRequired`, 
should we update to `articleMeta: PropTypes.oneOfType([null, corePropTypes.articleMetaForBookmark])` ?

packages/react-components/src/bookmark-widget/index.js
YuCJ: I think there might be  different states below for the bookmark widget status of an article:

```
  unknown: It has not checked the bookmark status yet
  isChecking: The request to check was make but has not gotten the response yet
  bookmarked: The article is bookmarked
  notBookmarked: The article is not bookmarked
  invalid: There's an error or there's no valid authentication info
```

The current code is a workaround to identify the `unknown` situation (by logging we've sent request for an article or not). We only want to send request to check bookmark when the status is `unknown`. We need to implement the status in the redux reducer and action. It can also prevent sending multiple requests to check the same article bookmark.

packages/react-components/src/bookmark-widget/index.js
```diff
@@ -80,50 +80,77 @@ const BookmarkImgDesktop = styled.div`
...
+          notBookmarked: The article is not bookmarked
+          invalid: There's an error or there's no valid authentication info
+        The current code is a workaround to identify the `unknown` situation (by logging we've sent request for an article or not).
+        We only want to send request to check bookmark when the status is `unknown`.
+     */
+    if (
+      articleSlug &&
+      !this.requestedBookmarkSlugs.some(value => value === articleSlug)
+    ) {
+      const hostFromWindow = getHostFromWindowLocation()
```
nickhsine: I am afraid that `hostFromWindow`can not solve SSR problem.
On server side rendering, we use [bookmark-widget data-loader](https://github.com/twreporter/twreporter-react/blob/master/src/routes.js#L163) to get the bookmark.
In bookmark-widget data-loader, we just give `host` a [certain value according to RELEASE_BRANCH](https://github.com/twreporter/twreporter-react/blob/master/src/data-loaders/bookmark-widget.js#L12-L17).
If we want to solve SSR problem,  we may need to get `host` in this file  just like how bookmark-widget data-loader get that.

packages/react-components/src/bookmark-widget/index.js
```diff
@@ -141,40 +168,59 @@ class BookmarkWidget extends React.PureComponent {
...
       )
     }
   }
 
+  checkIfThisArticleBookmarked() {
+    const { bookmark } = this.props
+    if (bookmark) {
+      const isSSR = typeof window === 'undefined'
+      const hostFromWindow = getHostFromWindowLocation()
+      if (!isSSR || bookmark.host !== hostFromWindow) {
```
nickhsine: `||` -> `&&` ?

packages/react-components/src/bookmark-widget/index.js
```diff
@@ -212,38 +258,43 @@ class BookmarkWidget extends React.PureComponent {
...
   const jwt = _.get(state, [reduxStatePropKeys.auth, 'accessToken'])
   const userID = _.get(state, [reduxStatePropKeys.auth, 'userInfo', 'user_id'])
   const isAuthed = _.get(state, [reduxStatePropKeys.auth, 'isAuthed'])
-  const bookmark = _.get(state, [reduxStatePropKeys.bookmarkWidget, 'bookmark'])
+  const bookmarkInStore = _.get(state, [
+    reduxStatePropKeys.bookmarkWidget,
+    'bookmark',
+  ])
+  const bookmarkForThisWidget =
+    currentSlug && currentSlug === _.get('bookmarkInStore', 'slug')
```
nickhsine: `_.get('bookmarkInStore'` -> `_.get(bookmarkInStore`

packages/react-components/src/bookmark-widget/index.js
nickhsine: `||` -> `&&` ?

packages/react-components/src/bookmark-widget/index.js
nickhsine: > I am afraid that `hostFromWindow`can not solve SSR problem.
> On server side rendering, we use [bookmark-widget data-loader](https://github.com/twreporter/twreporter-react/blob/master/src/routes.js#L163) to get the bookmark.
> In bookmark-widget data-loader, we just give `host` a [certain value according to RELEASE_BRANCH](https://github.com/twreporter/twreporter-react/blob/master/src/data-loaders/bookmark-widget.js#L12-L17).
> If we want to solve SSR problem, we may need to get `host` in this file just like how bookmark-widget data-loader get that.

Sorry for this false alarm.
I found the root cause, and I will reply on other comment.

packages/react-components/src/bookmark-widget/index.js
nickhsine: `||` -> `&&` ?

packages/react-components/src/bookmark-widget/index.js
nickhsine: `_.get('bookmarkInStore'` -> `_.get(bookmarkInStore`

packages/react-components/src/bookmark-widget/index.js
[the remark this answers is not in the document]
nickhsine: > If we remove `isRequired`,
> should we update to `articleMeta: PropTypes.oneOfType([null, corePropTypes.articleMetaForBookmark])` ?

The above comment is not correct.
Please ignore it.

packages/react-components/src/bookmark-widget/index.js
YuCJ: Fixed

packages/react-components/src/bookmark-widget/index.js
YuCJ: Fixed

YuCJ:
@nickhsine  Updated. Bump version to `v7.0.5-beta.5`. And it's tested okay with `twreporter-react` `RELEASE_BRANCH=master make start`

packages/react-components/src/bookmark-widget/index.js
```diff
@@ -90,45 +115,33 @@ class BookmarkWidget extends React.PureComponent {
...
-      const { jwt, userID, getSingleBookmark } = this.props
-      getSingleBookmark(
-        jwt,
-        userID,
-        this.getCurrentSlug(),
-        this.getCurrentHost()
+    /* TODO: Implement `status` for bookmark widget in redux reducer and action:
+      There should be different states below for the bookmark widget status of an article:
+        unknown: It has not checked the bookmark status yet
+        isChecking: The request to check was make but has not gotten the response yet
```
nickhsine: `make` -> `made`

packages/react-components/src/bookmark-widget/index.js
```diff
@@ -90,45 +115,33 @@ class BookmarkWidget extends React.PureComponent {
...
+        bookmarked: The article is bookmarked
+        notBookmarked: The article is not bookmarked
+        invalid: There's an error or there's no valid authentication info
+      The current code does not distinguish the `unknown`, `notBookmarked`, `isChecking`, and `invalid` situation.
+      The best result is that we only send request to check bookmark when the status is `unknown`.
+      But now we send request when there's no bookmark data for this component in the redux store when `componentDidMount`, no matter the reason of it.
+     */
```
nickhsine: 👍

PR #18, opened 2019-08-01
Fix several style problems

YuCJ:
- [x] Add margin between topic and title of leading if there's no subtitle. By changing margin-top of subtitle to the margin-bottom of topic.
- [x] Change image diff images order


nickhsine:
LGTM

nickhsine:
LGTM

PR #23, opened 2019-08-06
Update image components of `react-article-components`

YuCJ:
KanbanFlow task:

- [x] [Add `center-small` image](https://kanbanflow.com/t/c6hYaLuL)
- [x] [If no `src` is given (or is an empty string) to `img-with-placeholder`, it should render the placeholder as fallback](https://kanbanflow.com/t/LwERs21Z)

### DIFF
--- a/packages/react-article-components/src/components/body/index.js
+++ b/packages/react-article-components/src/components/body/index.js
@@ -1,3 +1,4 @@
+import alignmentConsts from '../../constants/element-alignment'
 import Annotation from './annotation'
 import Audio from './audio'
 import Blockquote from './blockquote'
@@ -9,8 +10,7 @@ import Image from './image'
 import ImageDiff from './image-diff'
 import Infobox from './infobox'
 import list from './list'
-import map from 'lodash/map'
-import mq from '@twreporter/core/lib/utils/media-query'
+import Multimedia from './multimedia'
 import Paragraph from './paragraph'
 import predefinedPropTypes from '../../constants/prop-types/body'
 import PropTypes from 'prop-types'
@@ -18,8 +18,14 @@ import React, { PureComponent } from 'react'
 import Slideshow from './slideshow'
 import styled, { css } from 'styled-components'
 import Youtube from './youtube'
+// @twreporter
+import mq from '@twreporter/core/lib/utils/media-query'
+// lodash
+import get from 'lodash/get'
+import map from 'lodash/map'
 
 const _ = {
+  get,
   map,
 }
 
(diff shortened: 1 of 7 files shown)


packages/react-article-components/src/components/body/index.js
```diff
@@ -180,7 +189,8 @@ const AlignRight = styled.div`
 
 function renderElement(data = {}) {
   const isCenterAligned =
-    data.alignment !== 'left' && data.alignment !== 'right'
+    data.alignment !== alignmentConsts.left &&
+    data.alignment !== alignmentConsts.right
```
nickhsine: change to `alignmentConsts .center` || `alignmentConsts.centerSmall` ?

packages/react-article-components/src/components/body/index.js
```diff
@@ -9,17 +9,22 @@ import Image from './image'
...
 import Paragraph from './paragraph'
 import predefinedPropTypes from '../../constants/prop-types/body'
 import PropTypes from 'prop-types'
 import React, { PureComponent } from 'react'
 import Slideshow from './slideshow'
 import styled, { css } from 'styled-components'
 import Youtube from './youtube'
+// @twreporter
+import mq from '@twreporter/core/lib/utils/media-query'
+import alignmentConsts from '@twreporter/core/lib/constants/element-alignment'
```
nickhsine: Do we only define `element-alignment` in `react-article-components/src/constants`? 
If we have other packages need this constant, then we move it to `core` package.
How do you think?

nickhsine:
Do we need to take `image.description` into account? 
The caption of `center-small` image will overlay with the following paragraph.

YuCJ:
> Do we need to take `image.description` into account?
> The caption of `center-small` image will overlay with the following paragraph.

I forgot the changing of `Image` width will change the position of `Caption`... I'm fixing this.

YuCJ:
Updated and tested okay on local

nickhsine:
We may need to rebase the latest commit:
`[twreporter-article-components] Update CHANGELOG.md` -> `[react-article-components] Update CHANGELOG.md`

packages/react-article-components/src/components/body/index.js
```diff
@@ -137,10 +143,34 @@ const StyledEmbedded = styled(Embedded)`
...
 `
 
-const StyledImageBlock = styled.div`
+const StyledExtendImageBlock = styled.div`
   ${extendWidthCSS};
 `
 
+const StyledLargeImageBlock = styled.div`
+  ${largeWidthCSS}
+  /* TODO: Consider to refactor image block component */
```
nickhsine: Which image block component do you want to refactor?

YuCJ: Just an intuition that overwrite the CSS beyond the component seems as the situation is an exception. But I think it's nice to see the position and style of small-center Caption is an exception for now. Because we are also not sure that what's the flexibility we need for the multimedia block component.

The comment seems too vague that I may forget what it means after a month... I think we will know which component should be refactor when a requirement occurs. So I removed the `TODO` comment.

PR #25, opened 2019-08-08
[react-article-components] prevent image flash while sliding between first and last slide

nickhsine:
### KanbanFlow Card
https://kanbanflow.com/t/q9SfBspy


packages/react-article-components/src/components/body/slideshow/index.js
```diff
@@ -599,45 +599,33 @@ export default class Slideshow extends PureComponent {
       images.slice(defaultCurIndex, slidesOffset)
     )
 
-    // since the items of imagesForSlicing would have the same id,
-    // we copy the images and append an index on their ids to avoid duplication.
     return _.map(imagesForSlicing, (img, index) => {
-      return _.merge({}, img, {
```
YuCJ: We can remove the `lodash/merge` in this module.

PR #30, opened 2019-08-14
[react-article-components]  implment table of contents

nickhsine:
This is a draft implementation.
Please help to check the APIs,
if there is no unclear designs, I will continue adding styles.

And this functionality will be moved to `@twreporter/react-components` once finished.


packages/react-article-components/src/components/table-of-contents/index.js
```diff
@@ -0,0 +1,379 @@
...
+    //  and the last highlight anchor is below the viewport
+    if (currentPosition === Waypoint.below) {
+      if (index <= 0) {
+        manager.highlightAnchor = null
+      } else {
+        manager.highlightAnchor = anchors[index - 1]
+      }
+    }
+  }
+
```
taylrj: Do we need to handle the other case that user scrolling down and the last highlight anchor is above the viewport in this function ?

nickhsine: I think this situation would be handled by `onEnter` of the last anchor.

nickhsine:
> If we add `Waypoint` to the anchor rather than to the section which contains other elements, it will not update the `manager.highlightAnchor` when the user "jump" to a position where no any anchor presents. (ex: When user refreshing the page or click on scroll bar of browser)

These are good questions.
Currently, we have two bugs:
1. when user refresh the page, the waypoint would not invoke any `onEnter` and `onLeave` functions. 
2. when user click scroll bar directly(in Firefox latest beta version), the `onEnter` and `onLeave` functions would not invoke in the correct order as we expect.

For bug 1., the easy solution would be `scrollTo(0, 0)` after page mounting, and so far the article page would do this for us. Hence, I am not going to fix this bug urgently.

For bug 2., I will spend some time to dig out `react-waypoint` source code to see what I can find.

nickhsine:
> For bug 2., I will spend some time to dig out `react-waypoint` source code to see what I can find.

The order of event triggering is the order we add into scroll event listener.
Once the scroll is too fast, all the anchors enter and leave the viewport at the same.
All events are emitting at the same time.
Event listener will notify all subscribers(anchors) one by one according to the order they are added into.
Therefore, the order is not like we expect.

nickhsine:
@YuCJ 
The latest commit fixes bug 2., please help to review it.

nickhsine:
@YuCJ  
[4dc5a65](https://github.com/twreporter/twreporter-npm-packages/pull/30/commits/4dc5a65235ce4ef2318f0e3af98e19903cc99f8f) and [07b5ce4](https://github.com/twreporter/twreporter-npm-packages/pull/30/commits/07b5ce4a936bbb42478901cdb4b6c608f65634a4) also fix bug 1.

packages/react-article-components/src/components/table-of-contents/index.js
```diff
@@ -55,13 +58,14 @@ class TOCManager {
...
    *  add instance of `Anchor` into array
+   *  and return a new array
    *  @method
    *  @param {Anchor} - an instance of `Anchor`
    *  @returns {undefined}
    */
   addAnchor(anchor) {
     if (anchor instanceof Anchor) {
-      this._anchors.push(anchor)
+      this._anchors = this._anchors.concat(anchor)
```
YuCJ: According to the comment added above, it should  `return this._anchors` ?

packages/react-article-components/src/components/table-of-contents/index.js
```diff
@@ -55,13 +58,14 @@ class TOCManager {
 
   /**
    *  add instance of `Anchor` into array
+   *  and return a new array
    *  @method
    *  @param {Anchor} - an instance of `Anchor`
    *  @returns {undefined}
```
YuCJ: According to the comment added above, it should be `@returns {Anchor[]}`?

nickhsine:
@YuCJ 
Please see the latest commits, including [283f79f](https://github.com/twreporter/twreporter-npm-packages/pull/30/commits/283f79f7f133a4e1040226cafd9276a64e0eaea3), [aee006a](https://github.com/twreporter/twreporter-npm-packages/pull/30/commits/aee006a2bc7a9a79791a169fe155d419cd835c0d), [7df7638](https://github.com/twreporter/twreporter-npm-packages/pull/30/commits/7df76386b11f8773751df24ce3a62dcea93504b7), [55f1c7f](https://github.com/twreporter/twreporter-npm-packages/pull/30/commits/55f1c7f59938961ca671b5b290f1b86d761cf26d) and [bc95bc4](https://github.com/twreporter/twreporter-npm-packages/pull/30/commits/bc95bc4cdbbfe6a4519ef4b7a1e17958f4f0f3df).
Those commits implements desktop version table of contents and extract reusable parts(named `this-file-will-be-moved-to-react-components`).

nickhsine:
@YuCJ 
Please review from [b9e5994](https://github.com/twreporter/twreporter-npm-packages/pull/30/commits/b9e5994350d887c3d4b2268cc9f22f60fa50dd27).

packages/react-article-components/src/components/aside/tools.js
```diff
@@ -117,6 +117,7 @@ const BackToTopicBlock = styled.div`
 
   ${mq.tabletAndBelow`
     &:after {
+      width: auto;
```
YuCJ: This will break the tooltip at Firefox

PR #37, opened 2019-09-17
[react-article-components] Fix embedded overflow problem

YuCJ:
KanbanFlow task: https://kanbanflow.com/t/w77swM2R


YuCJ:
> This change makes article horizontally scrollable if embed's width is bigger than viewport width(check demo page on mobile device).

I'll add `overflow: hidden` to prevent the horizontally scrolling as in article v1

nickhsine:
LGTM

PR #47, opened 2019-09-24
[react-article-components] fix embedded img not centered if not big enough

nickhsine:
### KanbanFlow Card
https://kanbanflow.com/t/fp8ce375


packages/react-article-components/src/components/body/embedded-code.js
```diff
@@ -15,6 +15,13 @@ const _ = {
 
 const Embedded = styled.div`
   position: relative;
+
+  img.img-responsive {
```
YuCJ: Since this is for special case only (image link), we can add comment about this

PR #48, opened 2019-09-24
[react-article-components] support empty paragrph

nickhsine:
### KanbanFlow Card
https://kanbanflow.com/t/fvB4VTjy


nickhsine:
Currently, there are many articles containing unnecessary empty paragraph blocks.
If we support empty paragraph, we have to clear unnecessary empty paragraphs manually,
and there are too many articles to change for editors. 
Hence, I will close this PR.

PR #50, opened 2019-09-26
[react-article-components] Add load-more and update relateds styles

YuCJ:
KanbanFlow task: https://kanbanflow.com/t/zWWBLo1j

Design: https://kanbanflow.com/t/E46FLGmc
> 此版本主要調整以下三部分：
> ① 新增「載入更多」按鈕
> ② Hover「載入更多」按鈕的樣式
> ③ 正在載入時卡片的照片樣式

### DIFF
--- a/packages/react-article-components/src/components/related/list.js
+++ b/packages/react-article-components/src/components/related/list.js
@@ -1,57 +1,137 @@
+import { sourceHanSansTC as fontWeight } from '@twreporter/core/lib/constants/font-weight'
 import Card from './card'
 import DynamicComponentsContext from '../../contexts/dynamic-components-context'
-import React from 'react'
-import get from 'lodash/get'
-import map from 'lodash/map'
+import mockup from './mockup'
 import mq from '@twreporter/core/lib/utils/media-query'
 import predefinedProps from '../../constants/prop-types/related'
-import styled from 'styled-components'
-import themeConst from '../../constants/theme'
+import React from 'react'
+import styled, { keyframes, css } from 'styled-components'
+import themeConsts from '../../constants/theme'
+// lodash
+import get from 'lodash/get'
+import map from 'lodash/map'
 
 const _ = {
   get,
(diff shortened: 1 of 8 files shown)


packages/react-article-components/src/components/related/list.js
```diff
@@ -59,39 +172,58 @@ export default class List extends React.PureComponent {
...
     const { data } = this.props
-    const { isMounted } = this.state
-
     if (data.length === 0) {
       return null
     }
+    const { showAll } = this.state
+    const cards = showAll
+      ? _.map(data, this.renderCard)
+      : _.map(data.slice(0, firstShowedLimit.desktopAndAbove), this.renderCard)
```
nickhsine: Problem:
We use `this.state.isMounted` to give each card `align-self: stretch` or `align-self: flex-start` style.
`stretch` to make sure card having the same height after mounting.
`flex-start` to make sure card having mouse hovering animation effects.

However, when the  time hidden cards `display: none` become `display: block`, `this.state.isMounted` is already `true`, which makes cards not have the same height.

The simple solution could be(for your reference)
```
// this.state.isMounted -> this.state.alignSelf
this.state = {
  showAll: false,
  alignSelf: 'stretch',
}

componentDidMount() {
  this.setState({
    alignSelf: 'flex-start',
  })
}

loadMore = () => {
  this.setState({
    showAll: true,
    alignSelf: 'stretch',
  }, () => {
    setTimeout(() => {
      this.setState({
        alignSelf: 'flex-start',
      })
    }, 500)
  })
}

renderCard = (item, i) => {
  ...
  <Item
     alignSelf={this.state.alignSelf}
     ...
  >
  ...
}
```

YuCJ: Fixed

YuCJ:
Updated

nickhsine:
👍 👍  👍

PR #60, opened 2019-10-18
Fix embedded script bug and remove `overflow:hidden`

YuCJ:


packages/react-article-components/src/components/body/embedded-code.js
```diff
@@ -101,15 +109,26 @@ export default class EmbeddedCode extends React.PureComponent {
...
-        node.appendChild(scriptEle)
+        try {
+          const scriptEle = document.createElement('script')
+          const attribs = script.attribs
+          const newAttribs = _groupAttribsDataset(attribs)
+          _.merge(scriptEle, newAttribs.notDataset, {
+            dataset: newAttribs.dataset,
+            text: script.text || '',
+            onload: dispatchLoadEvent,
+          })
```
nickhsine: Could we refactor this part as the following example?
```
const scriptEle = document.createElement('script')
const attribs = script.attribs
_.forOwn(attribs, (value, key) => {
  scriptEle.setAttribute(key, value)
})
scriptEle.text = script.text || ''
scriptEle.onload = dispatchLoadEvent
```
?

nickhsine: `_pickDatasetFromAttribs` is to handle old data camelized by v1 article page.
Since v2 article page won't pass camelized object into `EmbeddedCode`, we don't need this function anymore.

YuCJ: It's much more concise, thanks!

YuCJ: @nickhsine Updated

nickhsine:
LGTM

PR #68, opened 2019-11-06
[core] Use `request-origins.js` constants in `sigin-in-href.js` 

taylrj:
Problem description:

Since we are going to test on `next` release branch and `signInHref` in `sigin-in-href.js` doesn't have `next` branch as an attribute, font-end server crashes at [this line](https://github.com/twreporter/twreporter-npm-packages/blob/372554322a357cb2bef99f64586e376a13992b8b/packages/core/src/utils/sign-in-href.js#L38).

I propose to maintain only one constant file (`request-origins.js` ) to keep the origin information of different release branch.

Ref: [Kanban Flow Card](https://kanbanflow.com/t/BJCoy4EG)

### DIFF
--- a/packages/core/src/utils/sign-in-href.js
+++ b/packages/core/src/utils/sign-in-href.js
@@ -1,3 +1,4 @@
+import requestOrigins from '../constants/request-origins'
 import { getGlobalEnv } from './global-env'
 
 const { releaseBranch } = getGlobalEnv()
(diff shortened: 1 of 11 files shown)


packages/core/src/utils/sign-in-href.js
```diff
@@ -6,36 +7,12 @@ const signInSearchKeys = {
...
-    protocol: 'https',
-    host: 'accounts.twreporter.org',
-    pathname: signInPathname,
-    searchKeys: signInSearchKeys,
-  },
-}[releaseBranch]
 
 export function getSignInHref(destination = '') {
+  const signInOrigin =
+    requestOrigins.forServerSideRendering[releaseBranch].accounts
```
YuCJ: Use `forClientSideRendering` instead? Because `herf` to sign in is the link used at client side.

PR #69, opened 2019-11-07
fix: [react-components] bookmark-widget error on ssr

taylrj:
Fix ssr error [reported](https://console.cloud.google.com/errors/COC0tPq_4re0hAE?time=PT6H&project=coastal-run-106202&authuser=0) by Stackdriver

### DIFF
--- a/packages/react-components/src/bookmark-widget/index.js
+++ b/packages/react-components/src/bookmark-widget/index.js
@@ -114,10 +114,10 @@ class BookmarkWidget extends React.PureComponent {
 
   checkIfThisArticleBookmarked() {
     const { bookmark } = this.props
-    if (bookmark) {
-      const isClient = typeof window !== 'undefined'
+    const isClient = typeof window !== 'undefined'
+    if (bookmark && isClient) {
       const hostFromWindow = getHostFromWindowLocation()
-      if (isClient && _.get(bookmark, 'host') !== hostFromWindow) {
+      if (_.get(bookmark, 'host') !== hostFromWindow) {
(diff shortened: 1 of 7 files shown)


packages/react-components/src/bookmark-widget/index.js
```diff
@@ -116,7 +116,9 @@ class BookmarkWidget extends React.PureComponent {
     const { bookmark } = this.props
     if (bookmark) {
       const isClient = typeof window !== 'undefined'
-      const hostFromWindow = getHostFromWindowLocation()
+      const hostFromWindow = isClient
+        ? getHostFromWindowLocation()
+        : 'undefined'
```
YuCJ: We can just check `isClient` once:

```js
  checkIfThisArticleBookmarked() {
    const { bookmark } = this.props
    const isClient = typeof window !== 'undefined' 
    if (bookmark && isClient) {
      const hostFromWindow = getHostFromWindowLocation()
      if (_.get(bookmark, 'host') !== hostFromWindow) {
        /* Only check the consistency of `host` when client-side rendering. */
        console.warn(
          'Warning on checking bookmark status in `BookmarkWidget`:',
          'The `host` in the bookmark data is diffrent from the `host` in current `window`.',
          'host in bookmark:',
          bookmark.host,
          'host in `window.location`:',
          hostFromWindow
        )
      }
    }
    return Boolean(bookmark)
  }
```

taylrj: Updated.

PR #71, opened 2019-11-26
chore: add Lerna and bump version

taylrj:


packages/index-page/CHANGELOG.md
```diff
@@ -1,8 +1,15 @@
...
-## Unreleased
+All notable changes to this project will be documented in this file.
+See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
-## Release
+## [1.0.6](https://github.com/twreporter/twreporter-npm-packages/compare/@twreporter/index-page@1.0.6-rc.4...@twreporter/index-page@1.0.6) (2019-11-26)
+
+**Note:** Version bump only for package @twreporter/index-page
+
+## 1.0.6-rc.4 (2019-11-26)
```
YuCJ: Is there always an extra `rc.{n+1}` version when bump rc to stable version in Lerna?

taylrj: No, I tested the whole build and publish process with `Lerna` and published `v1.0.6-rc.4`.

So, if we bump current pre-release version to stable version, there would be no extra `rc.{n+1}` version.

PR #79, opened 2020-01-07
BREAKING CHANGE: redux: update redux actions and reducers

nickhsine:
BREAKING CHANGE: 
- resolves with success action rather than undefined
  - if data already exists, resolves with `{ type: types.dataAlreadyExists }`
  - if no more data, resolves with `{ type: types.noMoreItemsToFetch }`
  - if last request is processing, resolves with `{ type:
types.lastActionIsStillProcessing }`
- consolidate first layer properties of actions
  - action object structure should be
```
{
  type: actionType,
  payload: {
    // stuff should be handled by reducers
    // is put here.
  }
}
```
  - dispatch dataAlreadyExists, noMoreDataToFetch,  lastActionIsStillProcessing actions

- action will resolve with success action and reject with fail action.
Application code should catch action error, 
otherwise, it occurs Promise rejection not handled error.
- handle axios error by error-action-creators.axios
- isomorphic-fetch -> axios
- only contain informative properties of axios config
  - the following properties would be removed:
    - httpsAgent
    - httpAgent
    - socketPath
    - maxRedirects
    - validateStatus
    - onDownloadProgress
    - onUploadProgress
    - xsrfHeaderName
    - xsrfCookieName
    - adapter
    - paramsSerializer
    - paramsSerializer
- update unit tests
- getAccessToken action.payload.config is removed
- remove exported functions
  - failToReceiveAuthorCollection
  - failToFetchAuthorDetails
  - failToSearchAuthors

axios config reference:
https://github.com/axios/axios#request-config

### DIFF
--- a/packages/redux/src/actions/error-action-creators.js
+++ b/packages/redux/src/actions/error-action-creators.js
@@ -1,4 +1,5 @@
 import httpConsts from '../constants/http-protocol'
+
 // lodash
 import get from 'lodash/get'
 import merge from 'lodash/merge'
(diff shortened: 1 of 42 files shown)


nickhsine:
The main code change is in the first commit, 
@taylrj @YuCJ can just review that commit.
Other commits are to change unit test files.

packages/redux/src/actions/topics.js
```diff
@@ -98,12 +102,15 @@ function _fetchTopics(dispatch, origin, path, params, successActionType) {
...
+      const errorAction = errorActionCreators.axios(
+        error,
+        types.ERROR_TO_GET_TOPICS
+      )
       return dispatch({
-        type: types.ERROR_TO_GET_TOPICS,
+        type: errorAction.type,
         payload: {
-          error: e,
+          error: errorAction.payload,
```
taylrj: `payload: errorAction.payload` ?

nickhsine: I want to remain the same action payload as before.

nickhsine:
@YuCJ please help to review the last two commits.

packages/redux/src/actions/error-action-creators.js
```diff
@@ -50,15 +59,69 @@ function refineAxiosConfig(axiosConfig) {
...
+ *  @property {Object} request - Axios request
+ *  @property {Object} response- Axios response
+ *  @property {bool} isAxiosError
+ *  @property {string} message - Error message
+ *  @property {string} stack - Error stack
+ */
+
+/**
+ *  Refined Axios error object type definition
+ *  @typedef {Error} RefinedAxiosError
```
nickhsine: Do @YuCJ think we need to rename `RefinedAxiosError` ?
Is it clear for you to understand?

YuCJ: Naming is fine. If we want it more clear, maybe we can add some description like  "We only pick certain properties to be logged from default axios error because of the record size limitation (250kb) of the StackDriver."

packages/redux/src/store/create-store.js
```diff
@@ -38,8 +38,8 @@ export default async function createStore(
       window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__) ||
     compose
   const storeEnhancer = composeEnhancers(
-    applyMiddleware(thunkMiddleware.withExtraArgument({ httpClientWithToken })),
-    bindActionsToStore
+    bindActionsToStore,
+    applyMiddleware(thunkMiddleware.withExtraArgument({ httpClientWithToken }))
```
YuCJ: Can we document why the store enhancer returned by applyMiddleware should be the first enhancer here?

nickhsine: Sure, it will be great.

PR #87, opened 2020-01-17
chore: add ci/cd

YuCJ:


.circleci/config.yml
```diff
@@ -0,0 +1,103 @@
...
+            - node_modules
+          key: v1-dependencies-build-{{ checksum "yarn.lock" }}
+      - run:
+          name: "pick out lerna version commit"
+          command: |
+            VERSION_COMMIT_TITLE=$(node -e "console.log(require('./lerna.json').command.publish.message)")
+            COMMITS=$(git log "$GIT_BASE_REVISION".."$GIT_CURRENT_REVISION" --pretty="format:%h  %s")
+            echo "pushed commits:"
+            echo "$COMMITS"
+            VERSION_COMMITS_COUNT=$(echo $COMMITS | grep -cF "$VERSION_COMMIT_TITLE" || true)
```
babygoat: Why do we need `|| true`?

babygoat: Do we need `-F` option?

.circleci/config.yml
```diff
@@ -0,0 +1,103 @@
...
+            VERSION_COMMITS=$(echo $COMMITS | grep -F "$VERSION_COMMIT_TITLE" || true)
```
babygoat: VERSION_COMMITS=(${...}) to allow array compound assignment even empty. In this way you can avoid `|| true`

.circleci/config.yml
```diff
@@ -0,0 +1,103 @@
...
+            echo "version commits count: $VERSION_COMMITS_COUNT"
+            echo "version commits: $VERSION_COMMITS"
+            if [ $VERSION_COMMITS_COUNT -eq 0 ]; then
+              echo "There is no commit containing the release message."
+              VERSION_COMMIT_HASH=""
```
babygoat: Do we need to proceed if no release message?

packages/core/src/constants/categories.js
```diff
@@ -1,3 +1,5 @@
+/* void comment message for ci testing */
```
babygoat: Do we need to remove the test case?

packages/index-page/src/index.js
```diff
@@ -12,6 +12,7 @@ import ScrollFadein from './components/animations/scroll-fadein'
 import SeachBox from './components/search-box'
 import TopicsSection from './components/topics-section'
 
+/* void comment message for ci testing */
```
babygoat: Do we need to remove the test case?

.circleci/config.yml
YuCJ: > Why do we need `|| true` as the grep -c will print out 0 if non-matching?

If no `|| true` here, it will invoke `exit 1` and fail this ci build

babygoat: If multiple released commits, wouldn't it be better to error out early?

YuCJ:
Updated according to review. According to current workflow, we may run redundant overall test when bumping new version to master. May refactor it in the future.

YuCJ:
https://github.com/YuCJ/test-mono-repo-with-cicd
Playground project for simulating the workflow. Can try any action you want.

makefile
```diff
@@ -64,8 +64,10 @@ unlink:
 	yarn workspaces run unlink
 
 test:
-	@echo "$(P) Run tests"
-	NODE_ENV=test $(BIN_DIR)/jest
+	@echo "$(P) Run overall tests"
```
babygoat: Suggestion:
I might suggest renaming the overall tests to dependency validation to highlight the task explicitly.

babygoat: Besides, I might suggest removing the makefile test if the tests only check for integrity of the same file in each package. Provide the makefiles in each package but not intend for editing independently is not straightforward. We can have the changes after the discussion.

YuCJ: > Besides, I might suggest removing the makefile test if the tests only check for integrity of the same file in each package. Provide the makefiles in each package but not intend for editing independently is not straightforward. We can have the changes after the discussion.

Agree. I found that the better and common way is that the config files at root play the role like "base" config, and the config files in sub-folders (`pakckges/*/`) extend and overwrite the base configs.

.circleci/config.yml
```diff
@@ -1,108 +1,184 @@
...
+          requires:
+            - test
+          filters:
+            branches:
+              only: master
+      - publish:
+          name: publish_prerelease
+          prerelease: true
+          requires:
+            - test
```
babygoat: If workflow in each branch is 
PR:  test
Master:  test -> version(pre) -> publish (pre)
Release: test -> version(pre) -> publish (pre)
I might suggest using persist_to_workspace to avoid duplicated check_out install dependencies.

YuCJ: updated in latest commits

makefile
YuCJ: updated in latest commits

PR #90, opened 2020-02-05
fix: only new audio instance on client

YuCJ:

Howler.js should run on client only, not on server. (https://github.com/goldfire/howler.js#browser-compatibility)


nickhsine:
LGTM

PR #93, opened 2020-02-05
fix(react-article-components): update loadmore btn

YuCJ:
fix loadmore btn displaying logic.
not to show the btn when the amount of items under first show limit.


nickhsine:
LGTM

PR #94, opened 2020-02-05
Fix listing data reducer and update list component

YuCJ:
- fix(redux): fix mutating state in posts reducer
- feat(react-components): update list component: only wrap items with isFetching, do not include the header


nickhsine:
LGTM

PR #96, opened 2020-02-06
feat(core): add sign out href

YuCJ:


packages/core/src/utils/sign-out-href.js
```diff
@@ -0,0 +1,18 @@
...
+
+const signOutSearchKeys = {
+  destination: 'destination',
+}
+const signOutPathname = '/v2/auth/logout'
+const mainHref = requestOrigins.forClientSideRendering[releaseBranch].main
+
+export function getSignOutHref(destination = mainHref) {
+  const signOutOrigin = requestOrigins.forClientSideRendering[releaseBranch].api
+  const currentHrefSearch = destination
```
nickhsine: should we `encodeURIComponent` here?

PR #102, opened 2020-02-13
fix(universal-header): standalone header cannot work normally

nickhsine:
### Bug Description
The root cause is that the latest `@twreporter/redux/lib/store/createStore` return a `Promise` rather than a redux store instance.

### Solution
Decouple `@twreporter/redux`. 
Maintain redux store/actions/reducers by `@twreporter/universal-header` itself.

### DIFF
--- a/packages/universal-header/package.json
+++ b/packages/universal-header/package.json
@@ -19,7 +19,6 @@
   "license": "MIT",
   "dependencies": {
     "@twreporter/core": "^1.2.0",
-    "@twreporter/redux": "^6.0.1",
     "lodash": "^4.17.11",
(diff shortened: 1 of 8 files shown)


packages/universal-header/package.json
```diff
@@ -18,15 +18,16 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "@twreporter/core": "^1.2.0-rc.0",
-    "@twreporter/redux": "^6.0.1",
+    "@twreporter/core": "^1.1.2",
```
YuCJ: Should be `v1.2.0` after rebase on current `master`

PR #105, opened 2020-03-02
feat(errors): add new package errors

YuCJ:
Add custom errors to save the stack of errors. So we can add debug information when passing out the internal error.

### DIFF
--- a/packages/errors/README.md
+++ b/packages/errors/README.md
@@ -1,0 +1,298 @@
+# @twreporter/core
+
+[![NPM version](https://img.shields.io/npm/v/@twreporter/core.svg)](https://www.npmjs.com/package/@twreporter/errors)
+
+This package provides the helpers for handling errors in the [The Reporter](https://www.twreporter.org/) website.
+
+Inspired by [package `errors` for Go language](https://godoc.org/github.com/pkg/errors) and [the related article by Dave Cheney](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully).
+
+- [@twreporter/core](#twreportercore)
+  - [Use](#use)
+    - [errors.helpers.wrap](#errorshelperswrap)
+    - [errors.helpers.unwrap](#errorshelpersunwrap)
+    - [errors.helpers.cause](#errorshelperscause)
+    - [errors.helpers.printAll](#errorshelpersprintall)
+    - [errors.helpers.printOne](#errorshelpersprintone)
+    - [errors.helpers.annotateAxiosError](#errorshelpersannotateaxioserror)
+    - [errors.AnnotatingError](#errorsannotatingerror)
+      - [public properties](#public-properties)
+  - [Development](#development)
+    - [Dev](#dev)
+    - [Build](#build)
+    - [Publish](#publish)
+
+Reference:
+
+- [JavaScript Errors and Stack Traces in Depth](https://lucasfcosta.com/2017/02/17/JavaScript-Errors-and-Stack-Traces.html)
+- [V8 Stack trace API](https://v8.dev/docs/stack-trace-api)
+- [Don’t just check errors, handle them gracefully](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully)
+
+## Use
+
+### errors.helpers.wrap
+
+Annotate the error with `name`, `message` and `payload`. It will return a new error that is an instance of `errors.AnnotatingError`.
+
+See [source code and JSDoc](src/helpers/general.js#L4) for all parameters
+
+Example:
+
+```js
+import errors from '@twreporter/error'
+
+function doSomething() {
+  try {
+    const someArguments = {
+      /* ... */
+    }
+    invokeSomethingWithError(...someArguments)
+  } catch (error) {
+    console.error(
+      errors.helpers.wrap(
+        error,
+        'ErrorType',
+        'failed to invoke something',
+        { args: someArguments },
+        invokeSomethingWithError
+      )
+    )
+  }
+}
+```
+
+### errors.helpers.unwrap
+
+It will return the previous error in the chain that the input error belongs to.
+
+If there is no pointer of previous error saved in the input error, it will return `null`.
+
+It works like [the `Unwrap` method of errors in Go@^1.13](https://blog.golang.org/go1.13-errors#TOC_3.1.).
+
+See [source code and JSDoc](src/helpers/general.js#L46) for all parameters
+
+Example:
+
+```js
+import errors from '@twreporter/error'
+
+function doSomething() {
+  try {
+    const someArguments = {
+      /* ... */
+    }
+    invokeSomethingWithError(...someArguments)
+  } catch (error) {
+    /* list all error.name in errors chain */
+    let _error = error
+    const chain = []
+    while (_error) {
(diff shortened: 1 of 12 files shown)


packages/core/src/utils/error.js
```diff
@@ -0,0 +1,183 @@
...
+    }
+    this.name = 'ActionError'
+    this.payload = payload
+  }
+}
+
+/**
+ *
+ * @typedef MethodErrorPayload
+ * @property {string} method the method invoked. ex: '[lodash].get', '[ReactDOM].render()'
```
YuCJ: The `payload.method` here represents the current stack frame (function). It's used for tracking where this error from in the code.

It will be better to save the current stack frame automatically. I've consider to take the first line from `error.stack`, but the content of `error.stack` will be different according to JavaScript engine. (https://github.com/stacktracejs/error-stack-parser/blob/master/error-stack-parser.js#L27)

YuCJ: Or we do not have to save this info in the `error.payload` manually, since the info will be contained in the `error.stack` already. We just need to trim the stack info when we need to log all the errors in the `errorsStack` out.

packages/core/src/utils/error.js
```diff
@@ -0,0 +1,183 @@
...
+ */
+
+/**
+ *
+ *
+ * @export
+ * @class MethodError
+ * @extends {ErrorWithTrace}
+ */
+export class MethodError extends ErrorWithTrace {
```
YuCJ: Usage example:

```js
  const { MethodError } = require('twreporter/core/lib/utils/error');
  const sheets = require('googleapis').sheets({ version: 'v4' });
  function getPropertiesOfSheets() {
    const request = {
      spreadsheetId: this.spreadsheetId,
    };
    return sheets.spreadsheets
      .get(request)
      .then(res => {
        /* handle success */
      })
      .catch(error => {
        return Promise.reject(
          new MethodError('failed to get properties of sheets', {
            method: '[sheetsInstance].spreadsheets.get',
            args: [request],
          }, error)
        )
      })
```

YuCJ:
After discussion with @nickhsine and @babygoat, here's a draft of our JavaScript error handling principles:

1. Only handle (ex: log out) the error once. So in the most situations, a function encountering an error might throw the error out.

2. If there's any information not in the error and is helpful for debugging, wrap the error with an new error, and add these information to the new error (some developers call this "annotating" principle). Then throw this new error out.

3. Since the new error will have the pointer of its previous error, all errors related will form an errors chain. We can compose all the useful information in these errors. Since we have filter out the errors and the context of them that is useful, the information of new errors chain would be better than the built-in `error.stack` for debugging.

YuCJ:
The new helper functions are inspired by [package `errors` for Go language](https://godoc.org/github.com/pkg/errors) and [the related article by Dave Cheney](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully).

YuCJ:
The function frame information are retrieved automatically when creating an `Error` instance in JavaScript. So there's no need to distinguish the wrapper functions like `withStack` or `withMessage` here.

https://github.com/twreporter/go-api/pull/339

nickhsine:
LGTM
It's beautiful 🥇

packages/errors/README.md
```diff
@@ -0,0 +1,284 @@
...
+    at invokeSomething (/test.js:10:5)
+    at doSomething (/test.js:19:5)
+  payload undefined
+ErrorType: failed to invoke nested task
+    at invokeSomething (/test.js:12:26)
+  payload {"args":{"bar":"bar"}}
+ErrorType: failed to invoke something
+    at doSomething (/test.js:21:44)
+  payload {"args":{"foo":"foo"}}
+```
```
babygoat: Why do we need to duplicate nested stack frames of outer errors?

YuCJ: Some 3rd party libraries may throw deep-nested errors directly out to the consumer. I keep extra frames to reduce the probability of some important frames being trimmed. But after re-consideration, I found that it have two problems:

1. As you mentioned, if the cause error is not deep nested, it will provide duplicated stack data
2. Even I keep some frames in the stack, it cannot guarantee the most important frame (the interface to the consumer) is kept in the stack.

So I may not keep extra frames in the cause. And if we use any 3rd party libraries and the error does not provides sufficient information, we should add the information manually to the payload of annotating error.

packages/errors/src/annotating-error.js
```diff
@@ -0,0 +1,81 @@
...
+   *
+   * @returns {Error} the previous error in the chain in which this error locates
+   * @memberof AnnotatingError
+   * @public
+   */
+  unwrap() {
+    if (this._causer) {
+      return this._causer
+    }
+    return null
```
babygoat: Do we need branching here?

YuCJ: not necessary, but it's for make the API consistent with GO stderror

packages/errors/README.md
YuCJ: After discussion with @nickhsine. Since the stack is already generate (so the main cost of extra info is where to store the print out message), we think that the package could provide an option to print the untrimmed verbose information, and the default setting is not to save duplicate frames (for stackdriver memssage limit).

YuCJ:
@babygoat mentioned: need to test the performance of printing nested errors (compare printing native error)

YuCJ:
> @babygoat mentioned: need to test the performance of printing nested errors (compared with printing native error)

Although the new error wrapping method is about 4 times slower than the native error in average, but the average execution time in almost all cases is still under 1ms.

Wrap with native Error nested depth=10:

```js
function buildGoDeep(maxDepth, payload) {
  return function goDeep(depth = 0) {
    if (depth < maxDepth) {
      try {
        goDeep(depth + 1)
      } catch (error) {
        const err = new Error(`depth ${depth}`)
        err.payload = payload
        throw err
      }
    } else {
      throw new Error(`depth ${depth}`)
    }
  } 
}

const start = process.hrtime()
try {
  buildGoDeep(10, { ppp: 'pppp', ajiso: { asji: 'iaiso' } })(0)
} catch(err) {}
const end = process.hrtime(start)
const time = `${end[1] / 1000000}\n`
require('fs').appendFileSync('print-native-errors-depth-10.txt', time);
```

Execution time (run 20 times, different processes in sequence):

```
AVERAGE: 0.272 ms
MAX: 0.568 ms
MIN: 0.219 ms
```

Wrap with annotating errors nested depth=10:

```js
function buildGoDeep(maxDepth, payload) {
  return function goDeep(depth = 0) {
    if (depth < maxDepth) {
      try {
        goDeep(depth + 1)
      } catch (error) {
        throw errors.helpers.wrap(error, 'Error', `depth ${depth}`, payload)
      }
    } else {
      throw new Error(`depth ${depth}`)
    }
  } 
}

const start = process.hrtime()
try {
  buildGoDeep(10, { ppp: 'pppp', ajiso: { asji: 'iaiso' } })(0)
} catch(err) {}
const end = process.hrtime(start)
const time = `${end[1] / 1000000}\n`
require('fs').appendFileSync('print-annotating-errors-depth-10.txt', time);
```

Execution time (run 20 times, different processes in sequence):

```
AVERAGE: 0.904 ms
MAX: 1.769 ms
MIN: 0.739 ms
```

PR #118, opened 2020-05-29
update embedded component of react-article-components

YuCJ:
- dispatch window load event when each script is load => only dispatch window load event once after all scripts are load
- append script to DOM tree one by one => update DOM tree once with fragment


packages/react-article-components/src/components/body/embedded-code.js
```diff
@@ -96,9 +98,18 @@ export default class EmbeddedCode extends React.PureComponent {
           }
         })
         scriptEle.text = script.text || ''
-        scriptEle.onload = dispatchLoadEvent
-        node.appendChild(scriptEle)
+        const handleLoad = () => {
+          loadScriptsCount += 1
+          if (loadScriptsCount === scriptsCount) {
+            /* all scripts are load */
```
nickhsine, line 104: move venngage comments here?

PR #120, opened 2020-06-05
Redux refactoring according to go-api v2 doc

nickhsine:
### Things We Need to Know Before Reviewing
Notice: `post === article` in the following explanation.

- In article page, we need `post.relateds` and `post.topics.relateds` fields  to render the related articles section（文章頁底部的相關文章）. 
In short, related articles of a target article = related articles selected by editors + articles in the same topic.

- In topic page, we need `topic.relateds` to render the related articles section（專題頁下方的相關文章）.

- In mongo database, `post.relateds` and `post.topics.relateds` are an array of `ObjectID`s, we use `ObjectID` to refer to the entities.

- `FullPost` type is defined for article page.

- `FullTopic` type is defined for topic landing page.

- `Post` and `Topic` types are defined for listing. For listing articles or topics, we only need partial fields of `FullPost` or `FullTopic`. 

### Reason to Change
In the current (v1) backend service design, when we ask the backend service to return the a `full` article, it will embed articles (according to `ObjectID`s) into `relateds` and `topics.relateds` fields.

In the past, our articles are few, and there are not too many articles in the same topic.
Embedding is slow, which is a problem, but not a too big one.
However, `香港反送中` and `covid-19 疫情` topics are different, there are more than 30 articles in a single topic. 
It is too big to embed **ALL** related articles in the `full` article response. 
Right now, embedding would give backend service too much pressure and cause a series of requests timeout. 

In our backend service refactoring plans (v2), we don't embed `relateds` fields in the response anymore.
**Therefore, there are four affected components we need to update due to no embedded related posts:**
1. '最新專題' in index page
2. First topic in the topic listing page (`https://www.twreporter.org/topics?page=1`)
3. Related posts in the article page
4. Related posts in the topic landing page

### Notable Changes
- adopt go-api v2 response data structure change
- fetch related posts of a post on demand
- fetch related posts of a topic on demand
- fetch feature topic on demand
- refactor the way we fetch posts by list id (category id or tag id)

#### redux state
  - redesign `state.entities.posts` and `state.entities.topics` data structure
  - add `state.relatedPostsOf` (for related posts of a post or a topic)
  - add `state.featureTopic` (for index page and topic listing first page)

#### actions
  - remove `fetchInfographicPostsOnIndexPage`
  - remove `fetchEditorPickedPosts`
  - remove  `fetchPhotographyPostsOnIndexPage`
  - remove `fetchTopicsOnIndexPage`
  - update `fetchAFullPost`
  - update `fetchAFullTopic`
  - update `fetchTopics`
  - remove `fetchListedPosts`
  - add `fetchRelatedPostsOfAnEntity`
  - add `fetchPostsByCategoryListId`
  - add `fetchPostsByTagListId`
  - remove `fetchCategoriesPostsOnIndexPage`
  - update `fetchIndexPageContent`
  
#### action-types: 
  - remove `GET_TOPICS_FOR_INDEX_PAGE `
  - remove `GET_PHOTOGRAPHY_POSTS_FOR_INDEX_PAGE `
  - remove `GET_INFOGRAPHIC_POSTS_FOR_INDEX_PAGE `
  - remove `START_TO_GET_POSTS`
  - remove `ERROR_TO_GET_POSTS `
  - remove `GET_EDITOR_PICKED_POSTS `
  - remove `ERROR_TO_GET_EDITOR_PICKED_POSTS`
  - `GET_CONTENT_FOR_INDEX_PAGE` -> `indexPage.read.success`
  - `START_TO_GET_INDEX_PAGE_CONTENT` -> `indexPage.read.request`
  - `ERROR_TO_GET_INDEX_PAGE_CONTENT` -> `indexPage.read.failure`
  - `GET_LISTED_POSTS` -> `postsByListId.read.success`
  - `ERROR_TO_GET_LISTED_POSTS ` -> `postsByListId.read.failure`
  - add `postsByListId.read.request`
  - add `postsByListId.read.noMore`
  - add `postsByListId.read. alreadyExists`
  - add `relatedPosts.read.request`
  - add `relatedPosts.read.success`
  - add `relatedPosts.read.failure`
  - add `relatedPosts.read.noMore`
  - add `featureTopic.read.request`
  - add `featureTopic.read.success`
  - add `featureTopic.read.failure`
  - add `featureTopic.read.noMore`

#### reducers:
  - add `feature-topic.js`
  - add `related-posts-of.js`
  - update `entities.js`
  - update `posts.js`
  - update `topics.js`
  - update `index-page.js`

### Nice to Read Along with Reviewing
This [PR](https://github.com/twreporter/go-api/pull/387) specifies the go-api v2 response data structure. 
Our new redux action and reducer is depending on it.


packages/redux/src/reducers/index.js
```diff
@@ -21,6 +21,268 @@ const _ = {
...
+
+/**
+ *  Post type definition
+ *  @typedef {Object} Post
+ *  @property {Category[]} categories
+ *  @property {Image} hero_image
+ *  @property {Image} leading_image_portrait
+ *  @property {Image} og_image
+ *  @property {Object[]} brief.api_data
+ *  @property {Object} brief
```
nickhsine: reorder `brief` and `brief.api_data`.

nickhsine:
> In our backend service refactoring plans, we don't embed `relateds` fields in the response anymore.

The alternative is we lazy load `relateds` in the frontend.

babygoat:
I suggest we have the performance tests to evaluate the improvement during defined RPS.
 
1. v1 endpoint with the proposed redux state
2. v2 endpoint (w/ relateds in post/topic) with the current redux store
3. v2 endpoint (w/o relateds in post/topic) with the proposed redux store.

nickhsine:
> I suggest we have the performance tests to evaluate the improvement during defined RPS.
> 
> 1. v1 endpoint with the proposed redux state
> 2. v2 endpoint (w/ relateds in post/topic) with the current redux store
> 3. v2 endpoint (w/o relateds in post/topic) with the proposed redux store.
> 
> If 2 to 3 does not have explicit improvement, we can evaluate whether to adopt it or keep as it.

Sure, we can have the performance tests after v2 endpoint ready.

packages/redux/src/reducers/index.js
```diff
@@ -21,19 +21,330 @@ const _ = {
...
+/**
+ *  Post type definition
+ *  @typedef {Object} Post
+ *  @property {Category[]} categories
+ *  @property {Image} hero_image
+ *  @property {Image} leading_image_portrait
+ *  @property {Image} og_image
+ *  @property {Slug} slug
+ *  @property {Tag[]} tags
+ *  @property {bool} full
```
babygoat, line 92: Do we need full property?

packages/redux/src/reducers/index.js
```diff
@@ -21,19 +21,330 @@ const _ = {
...
+ *  @property {string} title
+ */
+
+/**
+ *  Topic type definition
+ *  @typedef {Object} Topic
+ *  @property {Image} leading_image
+ *  @property {Image} og_image
+ *  @property {Slug} slug
+ *  @property {bool} full
```
babygoat, line 108: Same above.

packages/redux/src/reducers/index.js
```diff
@@ -21,19 +21,330 @@ const _ = {
...
+ *  @property {string} updated_at
+ */
+
+/**
+ *  FullTopic type definition
+ *  @typedef {Object} FullTopic
+ *  @property {Image} leading_image
+ *  @property {Image} og_image
+ *  @property {Slug} slug
+ *  @property {bool} full
```
babygoat, line 122: Same above

packages/redux/src/reducers/index.js
```diff
@@ -21,19 +21,330 @@ const _ = {
...
+ */
+
+/**
+ *  FullPost type definition
+ *  @typedef {Object} FullPost
+ *  @property {Category[]} categories
+ *  @property {Image} hero_image
+ *  @property {Image} leading_image_portrait
+ *  @property {Image} og_image
+ *  @property {Slug} slug
```
babygoat, line 151: Same above.

packages/redux/src/reducers/index.js
```diff
@@ -21,19 +21,330 @@ const _ = {
...
+ * @property {number} height
+ * @property {number} width
+ */
+
+/**
+ *  Image type definition
+ *  @typedef {Object} Image
+ *  @property {string} id
+ *  @property {string} description
+ *  @property {string} copyright
```
babygoat, line 47: Where does the property be used?

packages/redux/src/reducers/index.js
```diff
@@ -21,19 +21,330 @@ const _ = {
...
+ *  @property {string} id
+ *  @property {string} name
+ */
+
+/**
+ *  Category type definition
+ *  @typedef {Object} Category
+ *  @property {string} id
+ *  @property {string} name
+ *  @property {number} sort_order
```
babygoat, line 72: Where is the sort_order used?

packages/redux/src/reducers/index.js
```diff
@@ -21,19 +21,330 @@ const _ = {
...
+/**
+ *  Image type definition
+ *  @typedef {Object} Image
+ *  @property {string} id
+ *  @property {string} description
+ *  @property {string} copyright
+ *  @property {string} filetype
+ *  @property {number} height
+ *  @property {number} width
+ *  @property {string} url
```
babygoat, line 51: Do we need the original image info?

packages/redux/src/reducers/index.js
```diff
@@ -21,19 +21,330 @@ const _ = {
...
+ *  @property {string} title
+ *  @property {string} topic_name
+ *  @property {string} updated_at
+ *
+ *  @property {Object[]} description.api_data
+ *  @property {Object[]} team_description.api_data
+ *  @property {Object} description
+ *  @property {Object} leading_video
+ *  @property {Object} team_description
+ *  @property {ObjectID[]} relateds - slugs of related posts
```
babygoat, line 134: objectID

babygoat:
To summarize the response requirement, we could still leverage the meta/full fetch.

Post
```
post.id
post.style 
post.slug 
post.leading_image_portrait
post.hero_image 
post.og_image
post.og_description
post.title
post.subtitle
post.categories
post.published_date
``` 
Full
```
post.brief
post.categories
post.content
post.copyright
post.designers
post.engineers 
post.extend_byline
post.hero_image 
post.id
post.is_external
post.leading_image_portrait
post.leading_image_description
post.og_description
post.og_image
post.og_title
post.photographers
post.published_date
post.relateds([]ObjectID)
post.slug
post.style
post.subtitle
post.title
post.topics(topic_name, slug,  relateds(objectID))
post.updated_at
post.writters
```
Topic

meta

```
topic.id 
topic.leading_image
topic.og_description
topic.og_image 
topic.slug 
topic.title
topic.topic_name
topic.relateds ([]objectID)
```

Full

packages/redux/src/reducers/index.js
nickhsine: Is this comment a typo?
Or do you mean that comment should change from
`slugs of related posts` -> `ObjectIDs of related posts`?

packages/redux/src/reducers/index.js
nickhsine: It would be great if we have `full` in the response.
It's a good indicator to check if articles in redux state are ready for article page or not.

packages/redux/src/reducers/index.js
nickhsine: https://github.com/twreporter/twreporter-npm-packages/blob/master/packages/react-article-components/src/components/related/index.js#L187
and 
https://github.com/twreporter/twreporter-npm-packages/blob/master/packages/react-article-components/src/components/aside/metadata.js#L262

packages/redux/src/reducers/index.js
nickhsine: No, we don't need it.
I will prune other unused properties of `Image` type in next commits.

packages/redux/src/reducers/index.js
nickhsine: We don't use it for now. 
I will prune it in next commits.

packages/redux/src/reducers/index.js
nickhsine: Do I miss something or is this comment a typo?

packages/redux/src/reducers/index.js
babygoat: Why should we use this field as indicators instead of other fields in full post/topic only?

packages/redux/src/reducers/index.js
babygoat: Sorry. False alarm

packages/redux/src/reducers/index.js
babygoat: The latter one is what I referred.

packages/redux/src/reducers/index.js
nickhsine: `full` indicator is more straightforward than other fields.
We used to take `content` field as indicator, but it's implicit for understanding.

babygoat: I suppose the point is that the field is used directly as the indicator. From the perspective of api response,  the `full` field only indicates the response contains full post but not whether the response ready. It seemed to me that whether the response is ready or not should be obtained from other indicator like http response code or the `isFetching`.

nickhsine: Just like you said, `full` field is used to indicate whether the post response is full or not.

babygoat: What is the exact use case? Indicate the response is ready? Or response is full article?

babygoat: What is the exact use case? Indicate the response is ready? Or response is full article?

babygoat: What is the exact use case? Indicate the response is ready? Or response is full article?

nickhsine: check if response is full article or not.

babygoat: OK, then I can keep the field.

packages/redux/src/typedef.js
```diff
@@ -0,0 +1,269 @@
...
+ *  FullTopic type definition
+ *  @typedef {Object} FullTopic
+ *  @property {Image} leading_image
+ *  @property {Image} og_image
+ *  @property {Slug} slug
+ *  @property {bool} full
+ *  @property {string} id
+ *  @property {string} og_description
+ *  @property {string} title
+ *  @property {string} topic_name
```
nickhsine, line 99: `topic_name` -> `short_title`

packages/redux/src/typedef.js
```diff
@@ -0,0 +1,269 @@
...
+ *  Topic type definition
+ *  @typedef {Object} Topic
+ *  @property {Image} leading_image
+ *  @property {Image} og_image
+ *  @property {Slug} slug
+ *  @property {bool} full
+ *  @property {string} id
+ *  @property {string} og_description
+ *  @property {string} title
+ *  @property {string} topic_name
```
nickhsine, line 85: `topic_name` -> `short_title`

packages/redux/src/typedef.js
```diff
@@ -0,0 +1,269 @@
...
+ *  @property {string} updated_at
+ *
+ *  @property {Object[]} description.api_data
+ *  @property {Object[]} team_description.api_data
+ *  @property {Object} description
+ *  @property {Object} leading_video
+ *  @property {Object} team_description
```
nickhsine, line 106: move `team_description.api_data` below `team_description`
move `description.api_data` below `description`

packages/redux/src/reducers/__test__/feature-post.test.js
```diff
@@ -0,0 +1,203 @@
+/* global expect, test, describe */
```
YuCJ, line 1: the filename should be `__test__/feature-topic.test.js`?

packages/redux/src/reducers/feature-topic.js
```diff
@@ -0,0 +1,73 @@
+import types from '../constants/action-types'
```
YuCJ, line 1: This new reducer does not be included in `src/reducers/index.js`?

packages/redux/src/reducers/posts.js
```diff
@@ -42,57 +40,87 @@ export function post(state = {}, action = {}) {
...
-      const listID = _.get(action, 'payload.listID', '')
       const page = _.get(action, 'payload.page', 0)
-      const list = _.get(state, listID, {
+      const oldItems = _.get(state, [listId, 'items'], [])
+
+      // Make sure we declare
+      // `items` and `pages` as `undefined`.
+      // The reason we use `undefined` is because
+      // `lodash/merge` will not replace the old value if
+      // new value is `undefined`.
```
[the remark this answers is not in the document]
nickhsine: Declaring `pages={}` and `items=[]` would be better than `undefined`.
I will update the comments and codes.

nickhsine:
> Since there's a schema change the `entities` reducer, some of the data in the cache of users' browsers will not be used anymore. Because we use `lodash/merge` to create the next state in the `entities` reducer and in the [cache function](https://github.com/twreporter/twreporter-npm-packages/blob/master/packages/redux/src/utils/browser-storage.js#L128), I guess these old data will still be kept and never expired.
> 
> In a future improvement, maybe we can add a key-value pair in local storage to document the cache was created via which version of `@twreporter/redux`, and clear the cache if their is a breaking change.

I thought the old redux state would expire https://github.com/twreporter/twreporter-npm-packages/blob/0e1a1735d9e04ad6f9f3b4d274ba54f2579707dc/packages/redux/src/utils/browser-storage.js#L136

Do I miss something?

YuCJ:
> > Since there's a schema change the `entities` reducer, some of the data in the cache of users' browsers will not be used anymore. Because we use `lodash/merge` to create the next state in the `entities` reducer and in the [cache function](https://github.com/twreporter/twreporter-npm-packages/blob/master/packages/redux/src/utils/browser-storage.js#L128), I guess these old data will still be kept and never expired.
> > In a future improvement, maybe we can add a key-value pair in local storage to document the cache was created via which version of `@twreporter/redux`, and clear the cache if their is a breaking change.
> 
> I thought the old redux state would expire
> 
> https://github.com/twreporter/twreporter-npm-packages/blob/0e1a1735d9e04ad6f9f3b4d274ba54f2579707dc/packages/redux/src/utils/browser-storage.js#L136
> 
> Do I miss something?

https://github.com/twreporter/twreporter-npm-packages/blob/0e1a1735d9e04ad6f9f3b4d274ba54f2579707dc/packages/redux/src/utils/browser-storage.js#L156

The current logic of cache will update the expire time on each redux store update. So the cache will not expire if the user visit our website continuously. We will change this logic in another PR to follow the rule as http cache.

packages/redux/src/actions/topics.js
```diff
@@ -221,14 +221,14 @@ export function fetchFeatureTopic() {
...
-          return Promise.all([
+          return [
             topic,
             {
               data: {
                 records: [],
               },
             },
-          ])
+          ]
```
taylrj, line 231:
```
return [
  topic,
  {
    data: {
        data: {
            records: []
        }
    }
  }
]
```

?

PR #122, opened 2020-06-10
feature: added text and via to Twitter share link

fredchu:
First of all, huge thanks for this great website you built. 

As a Twitter user and a reader of The Reporter, it could be convenient for readers to share an article by just clicking the Twitter share icon and not doing anything else.

But surprisingly, I found the website only provides only the link to Twitter Web Intent API, so readers have to add the title manually, which is not convenient for readers who would like to share the article like me.

I luckily found the source code and decided to contribute. I hope this won't interrupt your work.

Thank you.

### DIFF
--- a/packages/react-article-components/src/components/aside/tools.js
+++ b/packages/react-article-components/src/components/aside/tools.js
@@ -202,7 +202,7 @@ function TwitterShareBT(props) {
     const currentURL = window.location.href
     const location =
       'https://twitter.com/intent/tweet?' +
-      `url=${encodeURIComponent(currentURL)}&text=${encodeURIComponent(document.title)}&via=tw_reporter_org`
+      `url=${encodeURIComponent(currentURL)}&text=${encodeURIComponent(document.title + ' #報導者')}`
 
     window.open(location, '_blank')
   }
--- a/packages/react-article-components/src/components/aside/tools.js
+++ b/packages/react-article-components/src/components/aside/tools.js
@@ -202,7 +202,7 @@ function TwitterShareBT(props) {
     const currentURL = window.location.href
     const location =
       'https://twitter.com/intent/tweet?' +
-      `url=${encodeURIComponent(currentURL)}`
+      `url=${encodeURIComponent(currentURL)}&text=${encodeURIComponent(document.title)}&via=tw_reporter_org`
 
     window.open(location, '_blank')
   }


nickhsine:
LGTM 👍 👍 
Thanks for contributing.

Could you help to amend the commit message,
from `feature: added text and via to Twitter share link` -> `fix(react-article-components):  added text and via to Twitter share link` ?

packages/react-article-components/src/components/aside/tools.js
```diff
@@ -202,7 +202,7 @@ function TwitterShareBT(props) {
     const currentURL = window.location.href
     const location =
       'https://twitter.com/intent/tweet?' +
-      `url=${encodeURIComponent(currentURL)}`
+      `url=${encodeURIComponent(currentURL)}&text=${encodeURIComponent(document.title)}&via=tw_reporter_org`
```
nickhsine, line 205: After discussing with our editors, they wish we could use hash tag (#報導者) rather than via (@tw_reporter_org).

Therefore, could we change 
```
&text=${encodeURIComponent(document.title)}&via=tw_reporter_org`
```
to  
```
&text=${encodeURIComponent(document.title + ' #報導者')}`
```
?

fredchu: Adding `#報導者` sounds good! 

Why not just adding `#報導者` without removing the `via`? like

```
&text=${encodeURIComponent(document.title + ' #報導者')}&via=tw_reporter_org`
```

For readers, they could check `@tw_reporter_org` out after seeing this mention.

fredchu:
update: git message updated to `fix(react-article-components): added text and via to Twitter share link` cc @nickhsine

packages/react-article-components/src/components/aside/tools.js
nickhsine: Just check this with our twitter maintainers.

If someone retweets with `@tw_reporter_org`, our twitter maintainers will receive the system notifications.
When there are lots of system notifications, it's difficult for them to maintain.
 
That's the reason we don't want to add `via` parameter.

fredchu: Understood, will update the code. 👍

nickhsine: Thanks for contributing 🥇

PR #138, opened 2020-09-01
fix: toc disappears after article body rerenders

nickhsine:
#### Bug description:
We create `TOCManager` instance in each render of `Body` component,
and pass that instance into every `TOC.React.Anchor` as props.

The problem shows up when we rerender `Body`.
When we rerender `Body`, there are several steps happen:
1. recreate a new instance of `TOCManager`
2. pass the tocManager instance to every `Anchor`
3. `Anchor` won't be re-initialized if `key` prop not change
4. Due to step 3., `Anchors` information won't be stored into the tocManager instance
5. `TOC.React.TableOfContents` will dissapears after rerendering.

#### Solution:
Create the instance of `TOCManager` only once.
Therefore, `Anchors` information won't disappear after `Body` rerenders.

### DIFF
--- a/packages/react-article-components/src/components/article-page.js
+++ b/packages/react-article-components/src/components/article-page.js
@@ -487,6 +487,7 @@ export default class Article extends PureComponent {
                 {metadataAndToolsJSX}
                 <ContentBlock>
                   <Body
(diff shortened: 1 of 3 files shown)


YuCJ:
LGTM 👍

PR #144, opened 2020-09-08
fix(react-article-components): provide h1 string as anchor id

taylrj:


nickhsine:
LGTM 
Thanks for fixing this bug.

PR #150, opened 2020-09-15
refactor: stop using lodash.merge in entities reducer

nickhsine:


taylrj:
Just want to confirm about the meaning of following commit message:

```
when we render Article page,
we pick the article object from the entities.
The article object will be different, using `===` to compare, if we use
`lodash.merge` in the reducer.
```

The `lodash.merge` causes `redux.entities.posts` and `redux.entities.topics` change all the time, since it always returns new object which is always considered as different objects by `redux` and it causes `Article` to re-render.

Using shallow copy aims to prevent `Article` re-render when the value of `redux.entities.posts` and `redux.entities.topics` objects stays the same (unchanged).

nickhsine:
> Just want to confirm about the meaning of following commit message:
> 
> ```
> when we render Article page,
> we pick the article object from the entities.
> The article object will be different, using `===` to compare, if we use
> `lodash.merge` in the reducer.
> ```
> 
> The `lodash.merge` causes `redux.entities.posts` and `redux.entities.topics` change all the time, since it always returns new object which is always considered as different objects by `redux` and it causes `Article` to re-render.
> 
> Using shallow copy aims to prevent `Article` re-render when the value of `redux.entities.posts` and `redux.entities.topics` objects stays the same (unchanged).

Yes, you explain it clearer than me.
I will update the commit log.

taylrj:
LGTM 👍

nickhsine:
```
    In entities reducer, we return `newState` by
    `lodash.merge({}, state, nextState)`.
    In this way, `lodash.merge` will deep clone `state` and `nextState` to
    `newState`.

    Because `newState` is a deep clone,
    it will trigger React component rerendering.

    Take `Article` page as example,
    when we render `Article` page,
    we will pick the object from `state.entities.posts.byId.${postId}`
    to render.
    Since `newState` is a deep clone, `state.entities.posts.byId.${postId}`
    will be different if entities reducer executed (even the post is not
    changed).
    And that will causes `Article` page rerender.
```
Will this commit log be better to understand?

taylrj:
Yes, it will be easier for understanding.

PR #155, opened 2020-10-22
refactor(redux): update src/actions/author-details.js

nickhsine:


nickhsine:
Updated.

packages/redux/src/constants/action-types.js
```diff
@@ -5,10 +5,19 @@ export default {
   AUTH_FAILURE: 'grant_authorization_failure',
   AUTH_CLEAR: 'clear_authorization',
 
+  // keep these three action types here for backward compatibility.
```
taylrj, line 8: How about adding key word like `deprecated` here as a reminder ?

taylrj:
LGTM 👍

PR #157, opened 2020-11-12
fix(redux): deduplicate related posts

nickhsine:
### Reason to Change
Editors report that we render duplicate related posts.


taylrj (approved):
LGTM 👍

PR #163, opened 2021-01-12
Add podcast list link to footer

taylrj:
kanban flow card: 

[story-39][task-1] 在 footer 「基金會新聞專區」上方新增一個超連結「Podcast節目列表」，連結到 [podcast 節目清單頁](https://www.twreporter.org/a/podcast-list)

### DIFF
--- a/packages/react-components/src/footer/content.js
+++ b/packages/react-components/src/footer/content.js
@@ -43,7 +43,6 @@ function getItemGroups(mainOrigin) {
         text: '影響力報告',
         link: `${mainOrigin}${entityPaths.article}impact-and-annual-report`,
         target: '_self',
-        newFlag: true,
       },
     ],
     [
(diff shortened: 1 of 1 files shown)


packages/react-components/src/footer/content.js
```diff
@@ -43,7 +43,6 @@ function getItemGroups(mainOrigin) {
         text: '影響力報告',
         link: `${mainOrigin}${entityPaths.article}impact-and-annual-report`,
         target: '_self',
-        newFlag: true,
```
babygoat, line 46: Does this flag change requested by marketing department?

taylrj: No, it is suggested to change by designer (Gina).

babygoat: OK, it would be better to double check with 行銷.

babygoat:
LGTM

packages/react-components/src/footer/content.js
taylrj: Thanks for the reminder.

I have confirmed with `儀君` who approves this change.

PR #166, opened 2021-01-18
feat(index-page): add podcast box section

taylrj:


packages/index-page/src/components/podcast-box-section.js
```diff
@@ -0,0 +1,206 @@
...
+import PropTypes from 'prop-types'
+import React from 'react'
+import app from '../constants/app'
+import styled from 'styled-components'
+import url from 'url'
+
+// TODO(taylrj): update to the topic landing page link
+const podcastTopicLandingPageLink = 'https://www.twreporter.org/a/podcast-list'
+
+class PodcastLandingPageLinkWithUtm extends React.PureComponent {
```
babygoat, line 13: Why do we define the `PodcastLandingPageLinkWithUtm` here instead of @twreporter/react-components? How does the FE structure these packages?

packages/index-page/src/components/podcast-box-section.js
```diff
@@ -0,0 +1,206 @@
...
+  render() {
+    const { children, utmMedium } = this.props
+    const { isClient } = this.state
+    let podcastLandingPageURL = podcastTopicLandingPageLink
+
+    if (isClient) {
+      // client side rendering
+      try {
+        const utmSource = this.props.utmSource || window.location.host
+        const utmCampaign = this.props.utmCampaign || window.location.pathname
```
babygoat, line 40: Why  `window.location variable` can be utilized as the default value? does that make sense?

packages/index-page/src/components/podcast-box-section.js
taylrj: Components in `@twreporter/react-components` are more reusable and generic than `@twreporter/index-page`.
Basically,  `@twreporter/index-page` consumes components in `@twreporter/react-components` and do more things about representation (e.g. wrap it and decorate it).

At first, I believed `PodcastLandingPageLinkWithUtm` will not be used elsewhere. But now, I'd like to move it into `@twreporter/react-components` to make it possible to be reused except for index-page.

packages/index-page/src/components/podcast-box-section.js
taylrj: This idea comes from [donation-link-with-utm](https://github.com/twreporter/twreporter-npm-packages/blob/master/packages/react-components/src/donation-link-with-utm.js) which cares about the donation that individual article contributes.

Therefore, `utmSource` could be `https://www.twreporter.org` and `utmCampaign` could be article slug (e.g. `/a/article-slug`, that's the reason `window.location` variable be utilized here.

In case of `PodcastLandingPageLinkWithUtm`, `window.location.pathname` can be saved since we don't use article slug as `utmCampaign`. 

Besides, we can still utilize `window.location.host` as  default value of `utmSource` to trace each page view comes from index page.

babygoat: Thanks for the explanation

packages/index-page/src/components/podcast-box-section.js
```diff
@@ -0,0 +1,206 @@
...
+        const urlObj = new URL(podcastLandingPageURL)
+
+        urlObj.searchParams.set('utm_source', utmSource)
+        urlObj.searchParams.set('utm_medium', utmMedium)
+        urlObj.searchParams.set('utm_campaign', utmCampaign)
+
+        podcastLandingPageURL = url.format(urlObj)
+      } catch (e) {
+        console.warn('Can not get podcast landing page url with utm param', e)
```
babygoat: Minor. What is the possible exception to catch of the statements above? IMO, since `podcastLandingPageURL` is defined constant and we set the `searchParams` through API, no exceptions will occur during the execution of the statements.

packages/index-page/src/components/podcast-box-section.js
taylrj: Updated.

packages/index-page/src/components/podcast-box-section.js
taylrj: The exception that might occur in some legacy browsers which does not support `URL` api.

packages/react-components/src/podcast-link-with-utm.js
```diff
@@ -0,0 +1,58 @@
...
+    isClient: false,
+  }
+
+  componentDidMount() {
+    this.setState({
+      isClient: true,
+    })
+  }
+
+  setSearchParams(originalUrl) {
```
babygoat, line 26: If the member function is a `setter` function, I would not expect a return value. Maybe u are looking for `getter`? Like `getLinkWithSearchParams`?

taylrj: Indeed, updated.

packages/react-components/src/donation-link-with-utm.js
```diff
@@ -23,27 +22,28 @@ export default class DonationLinkWithUtm extends React.PureComponent {
...
+    const { utmSource, utmMedium, utmCampaign } = this.props
+    try {
+      const params = {
+        utm_source: utmSource || window.location.host,
+        utm_medium: utmMedium,
+        utm_campaign: utmCampaign || window.location.pathname,
+      }
+      return linkWithParams(originalUrl, params)
+    } catch (e) {
+      console.warn('Can not get donation url with utm param', e)
```
babygoat, line 35: If the exception occurs, do we need to return the original URL as fallback?

taylrj: Yes we need. Updated.

packages/react-components/src/donation-link-with-utm.js
```diff
@@ -23,27 +22,29 @@ export default class DonationLinkWithUtm extends React.PureComponent {
...
+    }
+  }
+
   render() {
-    const { className, children, utmMedium } = this.props
+    const { className, children } = this.props
     const { isClient } = this.state
     let donationURL = externalLinks.donation
 
     if (isClient) {
```
babygoat, line 45: Last question, why do we only append the utm tags on client side rendering?

taylrj: Since we might utilize `window.location` object as the default value of utm search params, we need to append these tags on client side rendering as `window` is not `undefined`.

babygoat: What about those who visit the index page directly? Would they have the link without utm tags?

taylrj: We implements server side rendering on index page which also renders on client side (client-side rendering).
On server side, the link has no utm tags but they are appended to url on client side.

We can validate the statement above by following steps:
1. Disable javascript on `https://twreporter.org`, refresh.
2. Now, without client side rendering, the donation link has no search params.
3. Enable javascript on `https://twreporter.org`, refresh.
4. Now, the donation link has search params since it renders on client side.

babygoat: Can I conclude that we need to re-render the index page since we need the default to use `windows.location`?

taylrj: The index page is always re-rendered on client side.

And there are other operations performed on client side, for example, fetching `latest-topic-section` data.

I would say we'd like to append utm tags on client side rendering since we need the default to use `window.location`.

babygoat: OK, thanks for the explanation.  LGTM

babygoat:
LGTM

PR #186, opened 2021-05-11
refactor(react-components): update side bar

taylrj:
This patch adds a more declarative way to configure side bar.
That is, directly declare the anchors in JSX.

### DIFF
--- a/packages/react-components/src/side-bar/index.js
+++ b/packages/react-components/src/side-bar/index.js
@@ -1,8 +1,8 @@
 import PropTypes from 'prop-types'
 import React from 'react'
-import { Waypoint } from 'react-waypoint'
 import hoistStatics from 'hoist-non-react-statics'
 import smoothScroll from 'smoothscroll'
+import { Waypoint } from 'react-waypoint'
 // lodash
 import get from 'lodash/get'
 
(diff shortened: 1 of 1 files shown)


taylrj:
Updated.

packages/react-components/src/side-bar/index.js
```diff
@@ -177,6 +182,17 @@ export default function decorateSideBar(DecoratedComponent) {
         modules = [children]
       }
 
+      const anchors = modules.map(module => {
+        const anchorObj = {
+          id: _.get(module.props, 'anchorId'),
+          label: _.get(module.props, 'anchorLabel', ''),
+          show: _.get(module.props, 'showAnchor', true),
```
babygoat, line 189: minor. would it be more intuitive to configure the default as `false` and display the anchor explicitly?

```
<DecoratedSideBar>
  <AnchorWrapper anchorId="id-1" anchorLabel="displayAnchor"  showAnchor >
      <SectionOne />
  </AnchorWrapper>
  <AnchorWrapper anchorId="id-2" anchorLabel="hideAnchor"  >
      <SectionTwo />
  </AnchorWrapper> 
</DecoratedSideBar>

```

babygoat:
LGTM

packages/react-components/src/side-bar/index.js
taylrj: Updated.

PR #190, opened 2021-06-16
refactor(react-article-components): add id tag to the aside tool

babygoat:
This patch adds the id tags to the social share icons(fb/twitter/line)
so that non-dev colleagues can leverage the CSS selector on GTM to
selector the icons independently.


taylrj:
LGTM

PR #194, opened 2021-07-13
refactor(universal-header): modify desktop icon show/hide flow

Aylie-Chou:
[Summary]
bookmark icon only show when login & modify desktop icon show/hide flow

[Related story]
twreporter-73

[Scope]
universal-header

[Change]
Before change, Icons component will always render bookmark icon & determine to render login/logout inside itself.
After change, header container component will determine which icons to render, then pass it to Icons component.

### DIFF
--- a/packages/universal-header/src/components/icons.js
+++ b/packages/universal-header/src/components/icons.js
@@ -1,5 +1,6 @@
 import HeaderContext from '../contexts/header-context'
 import Link from './customized-link'
+import PropTypes from 'prop-types'
 import React from 'react'
 import SearchBox from './search-box'
 import fonts from '../constants/fonts'
@@ -10,6 +11,12 @@ import styled from 'styled-components'
 import themeUtils from '../utils/theme'
 // @twreporter
 import mq from '@twreporter/core/lib/utils/media-query'
+// lodash
+import map from 'lodash/map'
+
+const _ = {
+  map,
+}
 
 const styles = {
   iconContainerSize: 3, // em
@@ -97,6 +104,14 @@ const HideOnDesktop = styled(IconContainer)`
 `
 
 class Icons extends React.PureComponent {
+  static propTypes = {
+    services: PropTypes.array,
+  }
+
+  static defaultProps = {
+    services: [],
+  }
+
   constructor(props) {
     super(props)
     this.state = {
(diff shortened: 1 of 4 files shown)


taylrj:
Since [Conventional Commits Specification](https://www.conventionalcommits.org/en/v1.0.0/) is applied in this repo for [lerna](https://github.com/lerna/lerna) to determine which version to bump (major, minor, or patch) and update CHANGELOG.md by reading [types](https://github.com/twreporter/Wiki/blob/master/contributing/commit-message-conventions-js-ts.md#type) in commit titles, please help to update current commit messages to meet the specification.

packages/universal-header/src/components/icons.js
```diff
@@ -10,6 +11,12 @@ import styled from 'styled-components'
 import themeUtils from '../utils/theme'
 // @twreporter
 import mq from '@twreporter/core/lib/utils/media-query'
+// lodash
+import map from 'lodash/map'
+
+const _ = {
+  map,
```
taylrj, line 18: Minor.
When do we use lodash map and when to use [built-in ES6 map](https://github.com/twreporter/twreporter-npm-packages/pull/194/files#diff-5596652809cb3425ddd6f94f72f791bec8b9ab03a890a71aad77b0506fee4f37R101) ?

Aylie-Chou: I like lodash because it's more readable for me.
Since universal-header already has lodash module for dependency, I would prefer to use lodash in new function.

For existing functions, I think we can choose one of them (not to use both in the single function).
Either built-in or lodash method is okay.

In the previous case, how do we choose which to use ?

taylrj: I raised the question because I believe we can choose one of them instead of use both.

Both are ok for me, you can make the call 😉.

Aylie-Chou: update in commit 6e077f3.
use lodash map in header container.

Aylie-Chou:
add commit 6e077f3
- remove TWREPORTER-73 feature
- use lodash map in header container

packages/universal-header/src/containers/header.js
```diff
@@ -111,6 +111,7 @@ class Container extends React.PureComponent {
     })
 
     if (isAuthed) {
+      const bookmarkKey = serviceConst.serviceKeys.bookmarks
```
taylrj, line 114: Unused `bookmarkKey` ?

Aylie-Chou: fix in commit 086b95a.
bookmarkKey appear after rebase 😂

taylrj:
LGTM

PR #195, opened 2021-07-21
feat(universal-header): adjust desktop category style & update icons

Aylie-Chou:
- add util/icons.js to load icon file
- remove mobile/tablet style in desktop only component (header, icons, channel)
- new category style

related to [TWREPORTER-7](https://twreporter-org.atlassian.net/browse/TWREPORTER-7)


packages/universal-header/src/containers/header.js
```diff
@@ -193,9 +167,12 @@ class Container extends React.PureComponent {
         <MobileOnly>
           <MobileHeader menu={mobileMenu} {...passThrough} />
         </MobileOnly>
-        <NonMobileOnly>
+        <TabletOnly>
```
taylrj, line 170: Why do we need `<TabletOnly>` if its child (i.e. `<MobileHeader>`) has no difference with `<MobileOnly>` ?

taylrj (commented):
Should we remove `packages/universal-header/static/photography/.DS_Store`, `packages/universal-header/static/normal/.DS_Store` ?

packages/universal-header/src/components/icons.js
```diff
@@ -30,6 +30,12 @@ const IconsContainer = styled.div`
   `}
 `
 
+const ShowOnHover = styled.div`
+`
```
taylrj, line 34: How about moving the hover selector into `ShowOnHover` and `HideOnHover` ?

such as:

```suggestion
const ShowOnHover = styled.div`
  display: none;
  &:hover {
    display: block;
  }
`
```

Is it doable ?

packages/universal-header/src/components/channels.js
```diff
@@ -64,99 +54,104 @@ const dropDownMenuEffectCSS = css`
...
   display: flex;
   flex-direction: row;
   flex-wrap: nowrap;
   align-items: center;
   list-style-type: none;
-  margin: 0;
+  margin: auto;
+`
+
+const ShowOnHover = styled.div`
```
taylrj, line 101: How about moving the hover selector into `ShowOnHover` and `HideOnHover` ?

packages/universal-header/src/components/drop-down-menu.js
```diff
@@ -18,17 +18,9 @@ const _ = {
...
-  contentMaxWidth: 320, // px
-  containerPadding: {
-    desktop: [10, 0], // px
-  },
   itemPadding: {
-    mobile: [30, 27], // px
-    tablet: [30, 33], // px
-    desktop: [9, 22], // px
-    hd: [9, 46], // px
+    desktop: [12, 16], // px
```
taylrj, line 22: Do we still need to specify `desktop` if the style is shared among other devices ?

packages/universal-header/src/components/icons.js
Aylie-Chou: The reason to put :hover in their parent is that the hover to show/hide effect should be triggered even when its parent been hoverd.
I think current method would make code a little bit dazzled.
Please let me know if you come out a better way for it.

packages/universal-header/src/containers/header.js
Aylie-Chou: The child of TabletOnly will be different in the next PR XD
Should I combine MobileOnly & TabletOnly in this version ?

taylrj: No, just keep it as it is. I'll resolve this.

packages/universal-header/src/components/icons.js
taylrj: Our use case seems like [referring-to-other-components](https://styled-components.com/docs/advanced#referring-to-other-components).

Would it be acceptable to you ?

Aylie-Chou: Looks great.
update it in commit 7478483

packages/universal-header/src/components/drop-down-menu.js
Aylie-Chou: Remove it in commit 7478483.

taylrj:
LGTM

PR #202, opened 2021-08-11
feat(universal-header): add new hamburger menu

Aylie-Chou:
- add HamburderMenu component: slide down -> full page
- add HamburgerService component: service icons for hamburder menu
- fix style


packages/universal-header/src/components/mobile-header.js
```diff
@@ -79,21 +134,24 @@ export default class MobileHeader extends React.PureComponent {
   }
 
   render() {
-    const { channels, services, actions } = this.props
+    const { actions, menuChannels, menuServices, menuActions } = this.props
 
     const { toSlideDown } = this.state
```
taylrj, line 139: Since the modal is pop-up instead of sliding down after this change, could we rename the state ?  Such as `isMenuOpen` ?

packages/universal-header/src/components/hamburger-icons.js
```diff
@@ -0,0 +1,235 @@
...
+    callback()
+  }
+
+  return (
+    <IconContainer>
+      <HeaderContext.Consumer>
+        {({ releaseBranch, theme }) => {
+          const [LogOutIcon] = themeUtils.selectHamburgerServiceIcons(theme).logout
+          return (
+            <a
```
taylrj, line 116: Do we want to reuse `<Link>` component ?

If so, we could modify the `to` prop to optional and renders `<a>` with no hyper link.

packages/universal-header/src/components/hamburger-icons.js
```diff
@@ -0,0 +1,235 @@
...
+    }
+  }
+
+  return (
+    <HeaderContext.Consumer>
+      {({ theme }) => {
+        const { borderColor } = themeUtils.selectHamburgerServiceTheme(theme)
+        return (
+          <IconsContainer borderColor={borderColor}>
+            { _.map(services, service => _prepareIconJsx(service)) }
```
taylrj, line 217: The pattern that prepares icons jsx here (and above) is beautiful 🎆 .

packages/universal-header/src/components/hamburger-menu.js
```diff
@@ -0,0 +1,155 @@
...
+    tablet: [24, 0, 42, 0],
+  },
+}
+
+const MenuContainer = styled.div`
+  background-color: ${props => props.bgColor};
+  position: fixed;
+  height: 100vh;
+  width: 100vw;
+  z-index: 101;
```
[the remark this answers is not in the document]
Aylie-Chou: Bingo, there's a component with z-index 100 in twreporter-react.

packages/universal-header/src/components/hamburger-icons.js
```diff
@@ -0,0 +1,235 @@
...
+  callback: () => {},
+}
+
+const Logout = ({callback}) => {
+  const handleLogout = (releaseBranch) => {
+    const redirectURL = window.location.href
+    const query = querystring.stringify({
+      destination: redirectURL,
+    })
+    window.location = linkUtils.getLogoutLink(releaseBranch).to + '?' + query
```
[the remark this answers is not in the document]
Aylie-Chou: Current redirect is window.location.href.
When server side rendering, server cannot recognize window object.
Any idea about leaving redirectURL defined in client side render ?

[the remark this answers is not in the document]
taylrj: Yes, you are right. Current solution is the best one that I can come up with. 

Even if there might be possible ways existed to leave redirectURL only to client side render, it is too much that a utility function should do. 

I'm resolving this.

Aylie-Chou:
updated.

- fix hamburger modal scroll problem (use same solution as twreporter-react pop-up-panel component)
- rename toSlideDown to isMenuOpen
- refactor customize-link component to accept empty props
- fix ios onClick event may not fire if target element is not interactive element

taylrj:
After the decision has been made on whether to move the pop-up panel to `@twreporter/react-components`, please comment to let me know. I'll resume the review.

Aylie-Chou:
updated.

- add mobile-pop-up-modal component to @twreporter/react-components to deal with parent scrolling problem in mobile device
- use mobile-pop-up-modal in hamburger-menu component

packages/react-components/src/mobile-pop-up-modal.js
```diff
@@ -0,0 +1,120 @@
...
+
+  render() {
+    const { modalHeight, modalWidth, ...rest } = this.props
+
+    return (
+      <Modal
+        height={modalHeight}
+        width={modalWidth}
+        ref={node => {
+          this.panel = node
```
taylrj, line 113: Would we like to create ref by using `React.createRef()` API to replace writing callback ref ?

There would be a little performance enhancement base on [#caveats-with-callback-refs](https://reactjs.org/docs/refs-and-the-dom.html#caveats-with-callback-refs) and the [related issue](https://github.com/facebook/react/issues/9328).

packages/react-components/src/mobile-pop-up-modal.js
```diff
@@ -0,0 +1,120 @@
...
+      'touchmove',
+      this.handleTouchmoveWhenPanning,
+      {
+        passive: false,
+      }
+    )
+  }
+
+  componentWillUnmount() {
+    this.panel = null
```
taylrj, line 59: Either we decide to use callback ref or `createRef` API, [React will assign the ref object back to null when it unmounts](https://reactjs.org/docs/refs-and-the-dom.html#adding-a-ref-to-a-dom-element).

I suppose we wouldn't need to set it back to null when component will unmont.

packages/react-components/src/mobile-pop-up-modal.js
```diff
@@ -8,6 +8,7 @@ const Modal = styled.div`
   position: fixed;
   z-index: 999;
   overflow-y: scroll;
+  max-height: -webkit-fill-available;
```
taylrj, line 11: 🚀

taylrj: Add a good reference here for someone who needs it in the future. (provided by @Aylie-Chou )

https://allthingssmitty.com/2020/05/11/css-fix-for-100vh-in-mobile-webkit/

taylrj (approved):
LGTM

packages/react-components/src/mobile-pop-up-modal.js
Aylie-Chou: Exactly, I'll leave to react.

PR #205, opened 2021-08-31
chore(universal-header): add dependency @twreporter/react-component

Aylie-Chou:
- for the first time included dependency, we should update package.json manualy
- lerna will take further updates

### DIFF
--- a/packages/universal-header/package.json
+++ b/packages/universal-header/package.json
@@ -19,6 +19,7 @@
   "license": "MIT",
   "dependencies": {
     "@twreporter/core": "^1.2.1",
+    "@twreporter/react-components": "^8.4.4",
     "lodash": "^4.17.11",
     "prop-types": "^15.6.2",
     "querystring": "^0.2.0",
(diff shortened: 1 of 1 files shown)


packages/universal-header/package.json
```diff
@@ -19,6 +19,7 @@
   "license": "MIT",
   "dependencies": {
     "@twreporter/core": "^1.2.1",
+    "@twreporter/react-components": "^8.4.4",
```
taylrj, line 22: @twreporter/react-components@8.5.0-rc.0 ?

Aylie-Chou: Is the previous react-component update merge the reason to use 8.5.0-rc.0 instead of released version ?

taylrj: I think so.
Otherwise, the next pre-release version of universal-header(i.e. `2.2.0-rc.3`) wouldn't contains the desired version of the dependency (`@twreporter/react-components@8.5.0-rc.0`).

Does it seem reasonable to you ?

Aylie-Chou: updated!
Thanks for explanation.

taylrj:
LGTM

PR #206, opened 2021-09-17
feat(universal-header): add narrow type header & header animation

Aylie-Chou:
feat(universal-header): add narrow type header & header transform animation

- add { toUseHeader, hideHeader } in container/header context
- add utils/animation.js to handle animation keyframe
- handle animation use CSSTransitionGroup

### DIFF
--- a/packages/universal-header/src/components/action-button.js
+++ b/packages/universal-header/src/components/action-button.js
@@ -66,6 +66,9 @@ const ActionsContainer = styled.div`
 `
 
 const ActionContainer = styled.div`
+  transition: opacity 0.3s;
+  transition-delay: 400ms;
+  opacity: ${props => props.isActive ? 1 : 0};
   margin : ${props => arrayToCssShorthand(styles.itemMargin[props.direction])};
   width: 100%;
 
@@ -115,6 +118,7 @@ const ActionItem = styled.div`
 
 const ActionButtonItem = ({ action, direction, callback }) => {
   const actionKey = action.key
+  const isActive = action.active === undefined ? true : action.active
   const actionLabel = actionConst.actionLabels[actionKey]
   const actionLink = linkUtils.getActionLinks()[actionKey]
   return (
@@ -122,7 +126,12 @@ const ActionButtonItem = ({ action, direction, callback }) => {
       {({ theme }) => {
         const { color, bgColor, hoverBgColor } = themeUtils.selectActionButtonTheme(theme)
         return (
-          <ActionContainer color={color} direction={direction} onClick={callback}>
+          <ActionContainer
+            color={color}
+            direction={direction}
+            isActive={isActive}
+            onClick={callback}
+          >
             <Link {...actionLink}>
               <ActionItem
                 direction={direction}
@@ -141,7 +150,8 @@ const ActionButtonItem = ({ action, direction, callback }) => {
 
 ActionButtonItem.propTypes = {
   action: PropTypes.shape({
-    key: PropTypes.string
+    key: PropTypes.string,
+    active: PropTypes.boolean,
   }),
   direction: PropTypes.string,
   callback: PropTypes.func,
--- a/packages/universal-header/src/components/action-button.js
+++ b/packages/universal-header/src/components/action-button.js
@@ -82,6 +82,7 @@ const ActionContainer = styled.div`
   a:visited {
     color: ${props => props.color || colors.white};
     font-size: ${styles.fontSize.desktop};
+    font-weight: ${fonts.weight.bold};
     ${mq.tabletOnly`
       font-size: ${props => styles.fontSize.tablet[props.direction]};
     `}
@@ -116,15 +117,15 @@ const ActionItem = styled.div`
   `}
 `
 
-const ActionButtonItem = ({ action, direction, callback }) => {
+const ActionButtonItem = ({ action, direction, themeFunction, callback }) => {
   const actionKey = action.key
-  const isActive = action.active === undefined ? true : action.active
+  const isActive = action.active ?? true
   const actionLabel = actionConst.actionLabels[actionKey]
   const actionLink = linkUtils.getActionLinks()[actionKey]
   return (
     <HeaderContext.Consumer>
       {({ theme }) => {
-        const { color, bgColor, hoverBgColor } = themeUtils.selectActionButtonTheme(theme)
+        const { color, bgColor, hoverBgColor } = themeFunction(theme)
         return (
           <ActionContainer
             color={color}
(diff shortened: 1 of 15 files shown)


packages/universal-header/src/containers/header.js
```diff
@@ -8,25 +8,37 @@ import channelConst from '../constants/channels'
...
+  throttle,
 }
 
+const HIDE_HEADER_THRESHOLD = 46
+const TRANSFORM_DURATION = 700
+
+const stickyTop = css`
+  position: sticky;
+  top: 0;
+  z-index: 9;
```
taylrj, line 33: Do we need to add comment here for why to choose 9 as z-index ?

packages/universal-header/src/components/action-button.js
```diff
@@ -115,14 +118,20 @@ const ActionItem = styled.div`
 
 const ActionButtonItem = ({ action, direction, callback }) => {
   const actionKey = action.key
+  const isActive = action.active === undefined ? true : action.active
```
taylrj, line 121: Would `someUndeclaredVariable === undefined` throws a ReferenceError ?

Or could we give it a little tweak, like:

```
typeof action.active === 'undefined' 
```
or

```
const isActive = action.active || true
```

?

taylrj:
It seems there's a styling issue for window innerWidth larger than 1440px.

image

taylrj:
When scrolling slightly, sometimes it stuck in an infinite animation.

https://user-images.githubusercontent.com/6375655/134460111-d6d1c525-2f3d-4e74-80d1-c3fcaed7f756.mov

packages/universal-header/src/components/action-button.js
Aylie-Chou: `const isActive = action.active || true` cannot handle action.active false case.

I'd like to use nullish coalescing operator for clarity.
`const isActive = action.active ?? true`
In order to support nullish coalescing syntax, I upgrade babel to latest version.
Let me know if you have any consideration of the upgrade.

Aylie-Chou:
updated.

- add themeFunction props to ActionButton, Channel, Slogan component
- upgrade babel preset-env to support nullish coalescing syntax

packages/universal-header/src/containers/header.js
Aylie-Chou: After checking, there is another component in @twreporter/twreporter-react having z-index 100,
thus z-index here should be assigned larger than 100.
Currently I assign it 101, would it be better to assign it as 999 (or larger) if the header should always on top ?

packages/universal-header/src/components/action-button.js
taylrj: sure, sorry for the incorrect suggestion.

Aylie-Chou:
> When scrolling slightly, sometimes it seems to be stucked in an infinite animation.
> 
>  Screen.Recording.2021-09-23.at.1.42.11.PM.mov

This case is caused by header height change after transforming.
When scroll more than 20 px, following UI change will occur:
1. transform to narrow header
2. scroll top become less than 20 due to header height change
3. transform to wide header
4. scroll top become more than 20 due to header height change
... back to step 1

packages/universal-header/src/containers/header.js
taylrj: I think 999 or larger would be better since we don't need to specify which is 100 and like the comment said, the header should always on top.

What do you say ?

taylrj: There's an edge case: CTA notify banner

https://user-images.githubusercontent.com/6375655/134498431-e6285a55-c1ba-4ad1-9e55-b89f1311ade7.mov

Is it also an issue about z-index ?

Aylie-Chou:
updated.
update __getScrollState function login to avoid infinite header transforming, and move state determined login from __handleScroll to __getScrollState function.

packages/universal-header/src/containers/header.js
```diff
@@ -92,16 +92,17 @@ class Container extends React.PureComponent {
...
       hideHeader: false,
     }
-    this.handleScroll = this.__handleScroll.bind(this)
+    this.handleScroll = _.throttle(this.__handleScroll, 350).bind(this)
 
-    let currentY = 0
-    let readyY = 0;
-    let isTransformDone = false
-    let transformTimer = null
+    // Below parameters are used to calculate scroll transform status.
```
taylrj, line 98: Breaking the reference by setting the variables to null on componentWillUnmount for garbage collection ?

packages/universal-header/src/containers/header.js
```diff
@@ -30,7 +30,7 @@ const TRANSFORM_DURATION = 700
 const stickyTop = css`
   position: sticky;
   top: 0;
-  z-index: 9;
+  z-index: 101; // there is another component in @twreporter/twreporter-react having z-index 100
```
taylrj: Do we need to set the z-index to 999 according to today's discussion ?

packages/universal-header/src/components/header.js
```diff
@@ -33,7 +33,7 @@ const styles = {
   channelMaxWidth: {
     hd: 1320, // px
   },
-  channelBottomBorderWidth: [1, 0, 1, 0], // px
+  channelBottomBorderWidth: [1, 0, 0, 0], // px
```
taylrj, line 36: Is it requested from designer ?

I suppose the original one is consistent with mockup.

Screeenshot after this change: 
image

Aylie-Chou: After checking with designer, channel should have border bottom.
Thx.

Aylie-Chou:
updated.
- fix hamburger not close after clicking channel item
- fix header should not slide out when first transform to narrow one

packages/universal-header/src/containers/header.js
```diff
@@ -98,7 +98,7 @@ class Container extends React.PureComponent {
     // Below parameters are used to calculate scroll transform status.
     this.currentY = 0
     this.readyY = 0;
-    this.isTransforming = false
+    this.isTransformfasle = false
```
taylrj, line 101: this.isTransforming ?

packages/universal-header/src/containers/header.js
```diff
@@ -120,6 +120,7 @@ class Container extends React.PureComponent {
   }
 
   __getScrollState(scrollTop, scrollDirection) {
+    const isCurrentNarrow = this.state.toUseNarrow
```
taylrj, line 123: nice fix 👍

taylrj:
others LGTM.

After the discussion, merging first for test.
@taylrj  will create patch to address comments in this PR.

PR #208, opened 2021-09-23
refactor(errors): remain the first line of stack trace

nickhsine:
### Change Reason
If the stack lines of error are less than trimmedCount, the whole stack lines will be deleted.
This patch remains the first line of error stack.


taylrj:
LGTM, thanks @nickhsine !

PR #212, opened 2021-09-28
fix: address review comment & lint error

taylrj:
- 'isTransforming' typo
- assign variables to null on componentWillUnmount hook
- remove 'mergeTwoArraysInOrder' which is defined but never used


Aylie-Chou (approved):
LGTM
eslint fix 的 style 跟我們原本的 style 不太一樣ＸＤ

PR #213, opened 2021-09-28
fix(universal-header): fix active state should remain after transforming & eslint error

Aylie-Chou:
- add currentPathname state in component/header & pass setPathname as callback props to Channel
- fix eslint


taylrj:
LGGGGGGTM, neat and tidy solution 👍

PR #215, opened 2021-10-04
fix: fix ui defect

Aylie-Chou:
- react-components: fix mobile pop up modal position & eslint error
- universal-header: fix style & channel active logic


Aylie-Chou:
updated.

- narrow header border-bottom & transition
- avoid header transform multiple times when scrollTop near 20px
- close channel dropdown menu when header hide

taylrj (approved):
LGTM

hard work 🏋️‍♂️🧗‍♀️

PR #216, opened 2021-10-05
refactor: update topic label style for article page

taylrj:
Address [TWREPORTER-158](https://twreporter-org.atlassian.net/browse/TWREPORTER-158).

This patch updates topic label style for article page to increase margin
between header and topic label for normal article page.

### DIFF
--- a/packages/react-article-components/src/components/leading/normal.js
+++ b/packages/react-article-components/src/components/leading/normal.js
@@ -103,7 +103,7 @@ const FigCaption = styled.figcaption`
 
 const BackgroundBlock = styled.div`
   ${props => getBackgroundBlockStyles(props.theme.name)}
-  padding-top: 10px;
+  padding-top: 30px;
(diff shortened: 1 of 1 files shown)


packages/react-article-components/src/components/leading/normal.js
```diff
@@ -70,7 +70,7 @@ const Topic = styled.div`
   border-width: 1px;
   border-style: solid;
   padding: 10px;
-  margin-bottom: 15px;
+  margin: 20px 0 15px 0;
```
Aylie-Chou, line 73: It seems not all article has topic part.
How about modify padding-top in L106 ?

taylrj: Updated.

Aylie-Chou (approved):
LGTM

taylrj:
Clean rebase.

PR #217, opened 2021-10-05
fix(universal-header): fix ui defect

Aylie-Chou:
- icons border color
- z-index management in Header component
- channel dropdown menu font weight & height


Aylie-Chou:
updated.
- fix tablet slogan animation delay & duration

taylrj:
LGTM

PR #218, opened 2021-10-06
fix(react-components): fix ios not update vh value when toolbar hide/show

Aylie-Chou:
This patch fix ios hambuger menu not fit viewport height due to vh value not update when toolbar show/hide.


taylrj:
beautiful solution, LGTM 👍

PR #219, opened 2021-10-07
fix(universal-header): fix ui defect

Aylie-Chou:
- enlarge hamburger icon click range
- disable click when action button inactive
- prevent browser click effect

### DIFF
--- a/packages/universal-header/src/components/mobile-header.js
+++ b/packages/universal-header/src/components/mobile-header.js
@@ -23,6 +23,10 @@ const styles = {
     mobile: [24], // px
     tablet: [24, 30, 24, 50], // px
   },
+  hamburgerPadding: {
+    mobile: [24], // px
+    tablet: [24, 30], // px
+  },
   logoHeight: {
     mobile: 26, // px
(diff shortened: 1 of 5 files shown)


packages/universal-header/src/components/mobile-header.js
```diff
@@ -125,7 +129,12 @@ const HamburgerContainer = styled.div`
 
 const Hamburger = styled.div`
   display: flex;
-  cursor: pointer;
+  position: absolute;
+  right: 0;
```
taylrj, line 133: add `top: 0;` ?

packages/universal-header/src/components/mobile-header.js
```diff
@@ -125,7 +129,12 @@ const HamburgerContainer = styled.div`
 
 const Hamburger = styled.div`
   display: flex;
-  cursor: pointer;
+  position: absolute;
```
taylrj, line 132: If `Hamburger` is set to position absolute, do we need to set `position: relative;` to its parent `FlexBox` or add a new wrapper with `position: relative;` ?

PR #234, opened 2021-11-04
refactor: manage font weight & font family in @twreporter/core

Aylie-Chou:
- add @twreporter/core/lib/constants/font
- deprecate @twreporter/core/lib/constants/font-weight


taylrj (approved):
LGTM

packages/core/src/constants/font.js
```diff
@@ -0,0 +1,14 @@
...
+  normal: '400',
+  bold: '700',
+}
+
+const fontFamily = {
+  title: 'merriweather, source-han-serif-tc, serif',
+  default: 'rosario, source-han-sans-traditional, sans-serif',
+}
+
+export default { fontWeight, fontFamily }
```
taylrj, line 12: Minor. Do we need default export if our use cases seem to use named import ?

Aylie-Chou: Default export is not neccessary in current use cases.
Do you think we would need this kind of flexibility ?

taylrj: Yes, we could keep this flexibility for future use.

PR #248, opened 2022-03-03
fix: disappear index in print mode

duidae:
套用media print使文章左側的索引在print mode不會顯示

dev可測的文章: http://localhost:3000/a/child-health-care-save-their-lives

jira 219
https://twreporter-org.atlassian.net/jira/software/c/projects/TWREPORTER/boards/7?modal=detail&selectedIssue=TWREPORTER-219

### DIFF
--- a/packages/react-article-components/src/components/table-of-contents/index.js
+++ b/packages/react-article-components/src/components/table-of-contents/index.js
@@ -1,13 +1,22 @@
-import * as Styled from './styled'
-import PropTypes from 'prop-types'
 import React from 'react'
-import TOC from '@twreporter/react-components/lib/table-of-contents'
+import PropTypes from 'prop-types'
 import map from 'lodash/map'
+import styled from 'styled-components'
+import * as Styled from './styled'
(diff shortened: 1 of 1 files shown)


packages/react-article-components/src/components/table-of-contents/index.js
```diff
@@ -1,14 +1,23 @@
...
+import * as Styled from './styled'
+import TOC from '@twreporter/react-components/lib/table-of-contents'
 import Tab from '../../assets/table-of-contents/long-form-tab.svg'
 
 const _ = {
   map,
 }
 
+const TOCContainer = styled.div`
+  @media print {
```
Aylie-Chou, line 14: Would you consider to put @media print setting in upper layer so that children components can controll the appearence with class name only (no need to create container for @media print setting) ?

duidae: good idea, merge #247, #248 to https://github.com/twreporter/twreporter-npm-packages/pull/249

PR #255, opened 2022-03-14
fix(react-article-components): show back to topic when published

taylrj:


duidae (changes requested):
nice to have bug fixed, btw could we also have a minor nullity check for robustness? e.g.
`const backToTopic = **leadingProps &&** leadingProps.isTopicPublished ? leadingProps.topicHref : ''`
thanks!

taylrj:
Updated.

duidae (approved):
Great, thanks!

PR #258, opened 2022-03-22
fix: update line-break property to avoid unexpected line break

taylrj:
This patch set `line-break: anywhere;` in title of editor picks to avoid
unexpected line break caused by punctuation characters.

### DIFF
--- a/packages/index-page/src/components/editor-picks.js
+++ b/packages/index-page/src/components/editor-picks.js
@@ -172,6 +172,7 @@ const Title = styled.div`
   left: 50%;
   transform: translateX(-50%);
   overflow: hidden;
+  line-break: anywhere;
   ${finalMedia.desktop`
     width: ${props => (props.middle ? '450px' : '120px')};
   `}


duidae (approved):
tested & issue is gone 👍 

btw a little bit curious about the reason of controlling the title with counting characters (>24, showing '...') instead of using `text-overflow: ellipsis`?

taylrj:
It seems `text-overflow: ellipsis;` works well when we want to introduce ellipsis on the first line. However, we need to show ellipsis '...' not on the first line but elsewhere, `text-overflow: ellipsis;` does not work in this case.

Another solution mentioned in this [[ref](https://css-tricks.com/almanac/properties/l/line-clamp/)] uses `display: --webki-box;` in its method, which is an legacy property and was replaced by `display: flex` now.

taylrj:
I'll merge it first since it has been approved by @duidae.
But if you have further concerns or suggestions, we can have more discussions here 🤓.

PR #263, opened 2022-04-26
Jason/issue 274

duidae:
fix [jira #274](https://twreporter-org.atlassian.net/browse/TWREPORTER-274)

### DIFF
--- a/packages/redux/src/actions/topics.js
+++ b/packages/redux/src/actions/topics.js
@@ -255,13 +255,13 @@ export function fetchFeatureTopic(timeout = apiConfig.timeout) {
         // fetch feature topic's latest three related posts
         .then(topic => {
           const allRelatedIds = _.get(topic, 'relateds', [])
-          const lastThreeRelatedIds = Array.isArray(allRelatedIds)
+          const threeRelatedIds = Array.isArray(allRelatedIds)
             ? allRelatedIds.slice(0, 3)
             : []
 
-          if (lastThreeRelatedIds.length > 0) {
+          if (threeRelatedIds.length > 0) {
             const url = formURL(apiOrigin, `/v2/${apiEndpoints.posts}`, {
-              id: lastThreeRelatedIds,
+              id: threeRelatedIds,
             })
             return Promise.all([
               topic,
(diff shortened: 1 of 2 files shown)


packages/redux/src/actions/topics.js
```diff
@@ -271,7 +269,17 @@ export function fetchFeatureTopic(timeout = apiConfig.timeout) {
                 .get(url, {
                   timeout,
                 })
-                .then(res => _.get(res, 'data.data.records', [])),
+                .then(res => {
+                  const relatedPosts = _.get(res, 'data.data.records', [])
+
+                  // Ensure the order of returned related posts is the same with topic landing page's order,
+                  // since API endpoint /v2/posts will sort the posts automatically by `published_date` in descending order.
```
taylrj, line 276: Question: Why is it necessary that API endpoint `/v2/posts` sorts the posts by `published_date`?

duidae: summary:
1. mongo may return nature order when `published_date` is the same. so we need to sync the returned result with id list.

2. /v2/posts endpoints may take a few sorting orders like:
sortByPublishedDate/sortByDescending + sortByPublishedDate/sortByUpdatedAt/sortByDescending + sortByUpdatedAt
ref: https://github.com/twreporter/go-api/blob/0dd7acdec758a5a132c57ac4d1837a91fcdc6e75/internal/news/query.go#L161

3. There are other places may reference /v2/posts:
https://github.com/twreporter/twreporter-npm-packages/blob/22fae0fb2066346296824c74daf767b440b01bff/packages/redux/src/actions/posts.js#L215

packages/redux/src/actions/topics.js
```diff
@@ -271,7 +269,17 @@ export function fetchFeatureTopic(timeout = apiConfig.timeout) {
...
+                  return relatedPosts.length < 2
+                    ? relatedPosts
+                    : lastThreeRelatedIds.map(id =>
```
taylrj, line 279: Its a good idea to ensure the order by utilizing `lastThreeRelatedIds`. It resolves mongoDB returns uncertain  order when the sort params has duplicate results.

packages/redux/src/actions/topics.js
```diff
@@ -258,7 +256,7 @@ export function fetchFeatureTopic(timeout = apiConfig.timeout) {
         .then(topic => {
           const allRelatedIds = _.get(topic, 'relateds', [])
           const lastThreeRelatedIds = Array.isArray(allRelatedIds)
```
taylrj, line 258: It seems we need to change the name `lastThreeRelatedIds` to `firstThreeRelatedIds` since its meaning has been changed since this PR.

duidae: renamed. please have a look, thanks.

taylrj (approved):
LGTM

PR #267, opened 2022-05-06
refactor: update howler config for large audio files

taylrj:
Address [TWREPORTER-317](https://twreporter-org.atlassian.net/browse/TWREPORTER-317)

This patch updates howler config for large audio files so that users
don't have to wait for the full file to be downloaded and decoded
before playing.

Also, set 'preload' to 'metadata' to preload the file's metadata
to get its duration without download the entire file.

ref: 
- html5: https://github.com/goldfire/howler.js#html5-boolean-false
- preload: https://github.com/goldfire/howler.js#preload-booleanstring-true

### DIFF
--- a/packages/react-article-components/src/components/body/audio/audio-provider.js
+++ b/packages/react-article-components/src/components/body/audio/audio-provider.js
@@ -115,6 +115,8 @@ export default class AudioProvider extends PureComponent {
       src: this.props.src,
       autoplay: false,
       loop: false,
+      html5: true,
+      preload: 'metadata',
       onload: () => {
         this.setState({
           duration: this._sound.duration(),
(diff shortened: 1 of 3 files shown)


packages/react-article-components/src/components/body/audio/audio-provider.js
```diff
@@ -115,6 +115,8 @@ export default class AudioProvider extends PureComponent {
       src: this.props.src,
       autoplay: false,
       loop: false,
+      html5: true,
+      preload: 'metadata',
```
Aylie-Chou, line 119: Question: according to howler [v2.1.2 manual](https://github.com/goldfire/howler.js/tree/v2.1.2#preload-boolean-true), `preload` is only accept as `boolean`.
Would we need to upgrade howler version for this feature ?

taylrj: According to the release [notes](url), it seems that we can upgrade howler version to the latest.
Update in https://github.com/twreporter/twreporter-npm-packages/pull/267/commits/488a57104d233d9d3aab2dbc9812c3701a04ecf1.

Aylie-Chou (approved):
LGTM

PR #272, opened 2022-05-25
fix: vercel build fail problem & missing bookmark-add icon

Aylie-Chou:
This patch fix vercel build fail problem & use design system component (change icon in gcs)


duidae (approved):
LGTM

PR #281, opened 2022-06-13
fix: bookmark should fetch upstream when page mount

Aylie-Chou:
- [TWREPORTER-182](https://twreporter-org.atlassian.net/browse/TWREPORTER-182): fetch new bookmarks when component mount


taylrj (approved):
LGTM

PR #287, opened 2022-07-04
fix: fix tool bar defect

Aylie-Chou:
- line & twitter icon cannot click (TWREPORTER-376)
- add timeout for snack bar (TWREPORTER-375)


duidae (approved):
LGTM

PR #291, opened 2022-07-20
fix: add id for bookmark & fix icon releaseBranch

Aylie-Chou:
- TWREPORTER-372
  - add `add-bookmark` & `remove-bookmark` id in order to exclude remove bookmark behavior on GTM trigger
- TWREPORTER-398
  - add releaseBranch for desktop-tools icon
  - pop out snackbar asynchronously

### DIFF
--- a/packages/react-article-components/src/components/aside/mobile-tool-bar.js
+++ b/packages/react-article-components/src/components/aside/mobile-tool-bar.js
@@ -292,6 +292,7 @@ const BookmarkBlock = ({ articleMeta }) => {
   const renderIcon = (isBookmarked, addAction, removeAction) => {
     const iconType = isBookmarked ? 'saved' : 'add'
     const text = isBookmarked ? '已儲存' : '加入書籤'
+    const id = isBookmarked ? 'remove-bookmark' : 'add-bookmark'
     const handleClick = () => {
       const action = isBookmarked ? removeAction : addAction
       action()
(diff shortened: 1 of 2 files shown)


packages/react-article-components/src/components/aside/mobile-tool-bar.js
```diff
@@ -292,15 +292,18 @@ const BookmarkBlock = ({ articleMeta }) => {
...
-    const handleClick = () => {
+    const id = isBookmarked ? 'remove-bookmark' : 'add-bookmark'
+    const handleClick = async () => {
       const action = isBookmarked ? removeAction : addAction
-      action()
-      setTimeout(() => {
-        toastr(isBookmarked)
-      }, 500)
+      try {
+        await action()
```
duidae, line 299: could we add error handling for action? i.e. `action?.()`

duidae (changes requested):

packages/react-article-components/src/components/aside/mobile-tool-bar.js
Aylie-Chou: if `action()` been rejected, it would also be handled by `catch` function in L300.

duidae: oh i see, thanks.

duidae (approved):

PR #298, opened 2022-08-04
feat: add title bar component for design guideline

Aylie-Chou:
- [TWREPORTER-329](https://twreporter-org.atlassian.net/browse/TWREPORTER-329)
- add `title bar` component
  - there are two types of title bar: `bar` & `tab`

### DIFF
--- a/packages/react-components/src/title-bar/components/tab.js
+++ b/packages/react-components/src/title-bar/components/tab.js
@@ -1,4 +1,10 @@
-import React, { useState, useEffect, useRef, useContext } from 'react'
+import React, {
+  useState,
+  useEffect,
+  useRef,
+  useContext,
+  createContext,
+} from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import { ScrollMenu, VisibilityContext } from 'react-horizontal-scrolling-menu'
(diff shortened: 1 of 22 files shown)


packages/react-components/src/title-bar/components/tab.js
```diff
@@ -0,0 +1,137 @@
...
+    }
+  }, [ref, falseOnScrollEnd])
+
+  return ref
+}
+
+const MobileTab = ({ tabs = [], activeTabIndex = 0 }) => {
+  const [activeIndex, setActiveIndex] = useState(activeTabIndex)
+  const [showGradientMask, setShowGradientMask] = useState(false)
+  const ref = useScrollStatus(setShowGradientMask, setShowGradientMask)
```
duidae, line 104: Mostly looking good, not sure why we are sending the same prop `setShowGradientMask` twice here for  useScrollStatus?

Aylie-Chou: It's because I tried to make `useScrollStatus` more general though current version still not so perfect with its two parameter `useScrollStatus (onHasScroll, falseOnScrollEnd)`.
Do you have any hint for simplifing `useScrollStatus` ?

duidae: what about just use comments to distinguish `onHasScroll` and `falseOnScrollEnd` literally, at the moment they both refer to `setShowGradientMask` & designing an extra parameter seems a little bit over-engineering

Aylie-Chou: combine `onHasScroll` & `falseOnScrollEnd` into `setShowNext` ([13aad8e](https://github.com/twreporter/twreporter-npm-packages/pull/298/commits/13aad8ec84d702af53be830d1ecdb482c15efd3a))

duidae (approved):
LGTM

PR #302, opened 2022-08-17
feat: add feature toggle for new information architecture

duidae:


packages/core/src/constants/feature-flag.js
```diff
@@ -0,0 +1,5 @@
+const enableNewInfoArch = false
```
Aylie-Chou, line 1: Could we also export `enableNewInfoArch ` in order to use `import { enableNewInfoArch } from xxx` syntax ?

duidae: okok

duidae: fixed in https://github.com/twreporter/twreporter-npm-packages/pull/302/commits/32d1633844abd9e57217e7b959c74621a24f11fc

Aylie-Chou (approved):
LGTM

PR #304, opened 2022-08-24
[enhancement] add `useSnackBar` hook

Aylie-Chou:
- react-components
  - add useSnackBar hook (headless UI !!)
  - add useSnackBar story
- react-article-components
  - apply useSnackBar hook


taylrj (approved):
Nice work !!!!!

It highly increases the flexibility using the snack bar component.

To me, the most impressive part is we no longer need to create and pass ref to react hook only to control the style. Just return props and let its parent to worry about style stuff. It really decouples the logic and pure style. 

If possible we could consider to "headlessify" more UI components in the future. Just follow the footstep in this change.

PR #313, opened 2022-09-29
feat: add new header components with feature toggle & storybook

Aylie-Chou:
- @twreporter/core 
  - add new color set `colorOpacity`
- @twreporter/react-components
  - add new components for design guideline
    - card/dialog
    - button/text button
    - input/search bar
  - storybook would change story background color with corresbonding theme
- @twreporter/universal-header


duidae:
Mostly looking good, and 2 issues for discussion:

1. values of theme and size('normal', 'photography',..., 'S', 'L') are spreading in some places, better to enum them(like color definition) in case value change in the future:
Screen Shot 2022-10-05 at 11 35 34 AM

2. This is a great & rich & work and so we have a big PR, perhaps we could split the task to 2~3 smaller PRs, so that we can digest it smoothly. According to a cisco's study, we could have a better defect rate from code review when PR is ~500 lines of code
https://smartbear.com/learn/code-review/best-practices-for-peer-code-review/

thanks.

Aylie-Chou:
updated ([2cf0499](https://github.com/twreporter/twreporter-npm-packages/pull/313/commits/2cf04990d34e0f26c7aee5519fb120ce99c12776))

- add THEME & BRANCH related consts in @twreporter/core package
- add SIZE & TYPE consts in components' own consts file
- fix pill button border-width

Aylie-Chou:
> perhaps we could split the task to 2~3 smaller PRs next time, so that we can digest it smoothly. According to a cisco's study, we could have a better defect rate from code review when PR is ~500 lines of code

Totally agree with that. The implement of `feature toggle` need to copy old files thus the diff would be duplicated. Maybe we could split PR by components if they're not coupling with each other.

duidae (approved):
LGTM!

PR #316, opened 2022-10-21
fix: storybook missing config

dannnyliang:
The storybook can't show any content and continuously loading.
I found that the console in dev tool show the error message `Uncaught ReferenceError: INITIAL_VIEWPORTS is not defined`

Because `INITIAL_VIEWPORTS` is removed in #313, the config can't be loaded correctly


Aylie-Chou (approved):
LGTM !
Thanks for the fix PR.

PR #326, opened 2022-11-14
fix: update kids banner

duidae:
fix https://app.asana.com/0/1202896150264815/1203309142649553/f
companion PR: https://github.com/twreporter/twreporter-react/pull/2299

### DIFF
--- a/packages/index-page/src/components/junior-box-section.js
+++ b/packages/index-page/src/components/junior-box-section.js
@@ -18,20 +18,20 @@ const mockup = {
 const mobileContentWidthPct = (mockup.contentWidth / mockup.defaultWidth) * 100
 
 const Container = styled.div`
-  background-image: url(${app.assetsPath}/PodcastBox_Desktop.jpg);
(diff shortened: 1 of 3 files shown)


Aylie-Chou:
According to [gcs schema guide](https://www.notion.so/twreporter/Google-Cloud-Storage-Schema-6479054cb8a04539900efb92804614b2), the final position for assets would be under `/assets`, and every assets would need to support different environments (master, staging, production...).
How about to take some efforts for the data migration this time ?

duidae:
> According to [gcs schema guide](https://www.notion.so/twreporter/Google-Cloud-Storage-Schema-6479054cb8a04539900efb92804614b2), the final position for assets would be under `/assets`, and every assets would need to support different environments (master, staging, production...). How about to take some efforts for the data migration this time ?

fixed accordingly, please have a look, thanks.

Aylie-Chou (approved):
LGTM 👍

PR #347, opened 2023-01-03
fix: defects

Aylie-Chou:
This patch fix defects in this [ticket](https://app.asana.com/0/1202896150264815/1202942985879116/f) & add `@babel/plugin-proposal-optional-catch-binding` to support `try catch` syntax.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -37,5 +37,6 @@ module.exports = {
         },
       },
     ],
+    '@babel/plugin-proposal-optional-catch-binding',
   ],
 }
--- a/babel.config.js
+++ b/babel.config.js
@@ -37,6 +37,5 @@ module.exports = {
         },
       },
     ],
-    '@babel/plugin-proposal-optional-catch-binding',
   ],
 }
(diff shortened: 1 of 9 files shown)


duidae (changes requested):
mostly looking good, but why do we need this package since `@babel/preset-env` already includes this plugin? we could also use try-catch syntax in various subpackages already (react-components, react-article-components...etc)

Aylie-Chou:
> mostly looking good, but why do we need `@babel/plugin-proposal-optional-catch-binding` since `@babel/preset-env` already includes this plugin? we could also use try-catch syntax in various subpackages already (react-components, react-article-components...etc)

After confirm, we do not need to add this package in additional.
Reverted it at [`4322136`](https://github.com/twreporter/twreporter-npm-packages/pull/347/commits/4322136c88c38229ce8d2528c25e2c675c2ca5b7).

Aylie-Chou:
Update [`eba130d`](https://github.com/twreporter/twreporter-npm-packages/pull/347/commits/eba130d0d670462ac25f40f25106a364d65fb160) for ios 15.1 only problem.
In ios15.1, url bar would collaspse/uncollapse while scrolling hamburger menu, thus hambuger menu would have change to render shorter height than it need.
This commit add `padding-bottom` for reserved usrl bar height.

duidae (approved):

PR #354, opened 2023-02-02
fix: fix rename issue

duidae:
fix field rename issue to avoid block testing, storybook will be another PR

Screen Shot 2023-02-02 at 4 58 33 PM


Aylie-Chou (approved):
LGTM

PR #355, opened 2023-02-07
Jason/add storybook

duidae:
depend on PR https://github.com/twreporter/twreporter-npm-packages/pull/354


packages/react-article-components/src/constants/theme.js
```diff
@@ -1,3 +1,4 @@
+// TODO: remove this constant & use @twreporter/core/lib/constants/theme
```
Aylie-Chou, line 1: Question: would we address todo in this pr ?

duidae: there are tons of places depending on this constant file, better to separate to another PR

packages/react-components/src/text/link.js
```diff
@@ -0,0 +1,58 @@
...
+  display: inline-block;
+  font-size: 16px;
+  line-height: 1;
+  padding-left: 5px;
+
+  &:hover {
+    padding-bottom: 2px;
+    border-width: 0 0 1px 0;
+    border-style: solid;
+    border-color: ${props => getColorFromTheme(props.theme)};
```
Aylie-Chou, line 30: Could we pass the value of `getColorFromTheme ` into <Text>  if `border-color` shares same value with `color` in L20 invoid calling `getColorFromTheme` twice ?

packages/react-components/src/text/link.js
```diff
@@ -0,0 +1,58 @@
...
+        style={{ fontWeight: props.isBold ? 'bold' : 'normal' }}
+      >
+        {props.name}
+      </Text>
+    </Anchor>
+  )
+}
+
+TextLink.propTypes = {
+  path: PropTypes.string,
```
Aylie-Chou, line 52: Question: would `path` props need to be required ?

packages/react-components/src/text/link.js
```diff
@@ -0,0 +1,58 @@
...
+      return COLOR_ARTICLE.brown
+  }
+}
+
+const Text = styled.div`
+  color: ${props => getColorFromTheme(props.theme)};
+  display: inline-block;
+  font-size: 16px;
+  line-height: 1;
+  padding-left: 5px;
```
Aylie-Chou, line 24: As a shared component, I think it would be better to assign `padding` in where you use it.
We could accept `...props` or `className` as <TextLink> prop  to make it a styled-able component.

Aylie-Chou (changes requested):
Looks like a huge work to unify colors ! I would discuss the color issue with pm & designer for the futher work.
Mostly looking great, just some fine tune for the shared component in react-component.

packages/react-components/src/text/link.js
duidae: good point, fixed.

packages/react-components/src/text/link.js
duidae: fixed.

packages/react-components/src/text/link.js
duidae: fixed.

duidae:
> Looks like a huge work to unify colors ! I would discuss the color issue with pm & designer for the futher work. Mostly looking great, just some fine tune for the shared component in react-component.

comments are fixed as above, please have a look, thanks.

PR #357, opened 2023-02-15
fix: category politics and society path typo

Aylie-Chou:
This patch fix following defects:
- [`[defect] 首頁議題區塊只顯示 7 個區塊，少了政治社會`](https://app.asana.com/0/0/1203964075873158/f)
- [`[defect] 主網站：分類頁的 title bar 與下面的文章 margin: tablet 應為 32px，mobile 應為 24px`](https://app.asana.com/0/0/1203964971392538/f)
- [`[defect] 主網站：最新頁的 title bar 與下面的文章 margin: tablet 應為 32px，mobile 應為 24px`](https://app.asana.com/0/0/1203964971392530/f)
- [`[defect] 最新頁：在最新頁點選其他時事時，tab bar 的最新 tab 應該要 active state（變成紅色）`](https://app.asana.com/0/0/1203964971392539/f)

### DIFF
--- a/packages/core/src/constants/category-set.js
+++ b/packages/core/src/constants/category-set.js
@@ -1,7 +1,7 @@
 const categoryPath = {
   world: 'world',
   humanrights: 'humanrights',
-  politicsAndSociety: 'politics-and-society',
+  politicsAndSociety: 'politics_and_society',
   health: 'health',
   environment: 'environment',
   econ: 'econ',
--- a/packages/core/src/constants/category-set.js
+++ b/packages/core/src/constants/category-set.js
@@ -1,7 +1,7 @@
 const categoryPath = {
   world: 'world',
   humanrights: 'humanrights',
-  politicsAndSociety: 'politics_and_society',
+  politicsAndSociety: 'politics-and-society',
   health: 'health',
   environment: 'environment',
   econ: 'econ',
--- a/packages/universal-header/src/utils/links.js
+++ b/packages/universal-header/src/utils/links.js
@@ -79,12 +79,7 @@ export const checkReferrer = (
 }
 
 export const checkPathnameParent = (pathname = '', parent = '') => {
-  try {
-    const isMatch = _.indexOf(_.split(pathname, '/'), parent) === 1
-    return isMatch
-  } catch (err) {
-    return false
-  }
+  return _.indexOf(_.split(pathname, '/'), parent) === 1
 }
 
 export const getCategoryLink = (
--- a/packages/universal-header/src/utils/links.js
+++ b/packages/universal-header/src/utils/links.js
@@ -80,7 +80,7 @@ export const checkReferrer = (
 
 export const checkPathnameParent = (pathname = '', parent = '') => {
   try {
-    const isMatch = _.split(pathname, '/')[1] === parent
+    const isMatch = _.indexOf(_.split(pathname, '/'), parent) === 1
     return isMatch
   } catch (err) {
     return false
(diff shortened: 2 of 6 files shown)


packages/core/src/constants/category-set.js
```diff
@@ -1,7 +1,7 @@
 const categoryPath = {
   world: 'world',
   humanrights: 'humanrights',
-  politicsAndSociety: 'politics-and-society',
```
duidae, line 4: also found this string in src/containers/Home.js of `twreporter-react`

duidae (changes requested):

packages/core/src/constants/category-set.js
Aylie-Chou: updated ([ec777df](https://github.com/twreporter/twreporter-npm-packages/pull/357/commits/ec777df0caf96f58a0c678db5fff667c6a8481ae))
frontend key name is different from backend api returned key name, thus address key name transform on interface between fe & be (redux)

packages/universal-header/src/utils/links.js
```diff
@@ -78,6 +78,15 @@ export const checkReferrer = (
   }
 }
 
+export const checkPathnameParent = (pathname = '', parent = '') => {
+  try {
```
duidae, line 82: I think it would better to check index of array ahead to avoid index out of bound(which could really happen) & using try-catch for error handling

duidae (changes requested):

packages/universal-header/src/utils/links.js
Aylie-Chou: The index out of bound case would also catch by try-catch syntax.
The reason not to check out of bound case is that we do not need to split different kind of errors here. (all of the errors should return false)

duidae: out of bound這種error通常需要被嚴肅對待因為他是有可能造成app crash, 有點好奇用try-catch這樣的招式是從何而來和有何益處, 而且應該是非不得已才使用, 建議還是檢查array index比較正式一點

Aylie-Chou: updated. ([68b7ff5](https://github.com/twreporter/twreporter-npm-packages/pull/357/commits/68b7ff54c4c706a6b6eacc7869376f117e46928a))

After checking, I found that javascript arrya is formed by object, thus js array do not have out of bound case, it would return `undefined` for it, so that there's no error throwing for us to catch.
To prevent using more memory here, I change the function to check the index, please have a look.

duidae: then the whole try-catch thing is not necessary, why dont we just
```
const tokens = _.split(pathname, '/')
return tokens?.length >= 2 && tokens[1] === parent
```

duidae (changes requested):

packages/universal-header/src/utils/links.js
Aylie-Chou: There's two reason for it.
1. we also need to match and return true for  '/latest/xxx' case
2. Try catch here means a ensure for any other error cases. Ex. input parameter type change

duidae: for 1. we can evaluate the cost of calling `both split() & indexOf()` vs calling only `split()`, leave to you decide.

Aylie-Chou: updated ([a0c80a7](https://github.com/twreporter/twreporter-npm-packages/pull/357/commits/a0c80a7255ad145884acd1a88eaedc702f7e27c3))

- remove unnecessary try-catch (lodash would handle error here)

duidae (approved):

PR #358, opened 2023-02-15
fix: fix category link issue & add error handling

duidae:
fix
[[defect] 點擊文章頁 meta data 的分類/子分類時，無法正確跳轉到對應頁面](https://app.asana.com/0/1202961977949746/1203964075873151/f)
[[defect] keystone 設定的分類組合未完整顯示於文章頁](https://app.asana.com/0/1202961977949746/1203964075873152/f)

### DIFF
--- a/packages/react-article-components/src/components/aside/metadata.js
+++ b/packages/react-article-components/src/components/aside/metadata.js
@@ -15,6 +15,7 @@ import { idToPathSegment } from '../../constants/category'
 import mq from '@twreporter/core/lib/utils/media-query'
 import { ARTICLE_THEME } from '@twreporter/core/lib/constants/theme'
 import { COLOR_ARTICLE } from '@twreporter/core/lib/constants/color'
+import { GET_CATEGORY_PATH_FROM_ID } from '@twreporter/core/lib/constants/category-set'
 import { ENABLE_NEW_INFO_ARCH } from '@twreporter/core/lib/constants/feature-flag'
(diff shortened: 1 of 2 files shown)


packages/react-article-components/src/components/aside/metadata.js
```diff
@@ -293,14 +294,15 @@ const CategorySet = props => {
         </CategorySetFlex>
       )
     }
-    return (
+    const categoryPath = GET_CATEGORY_PATH_FROM_ID(set?.category?.id)
```
Aylie-Chou, line 297: The returned `post.category_set` data from go-api whould fetch category & subcategory name.
If there's not other consideration,  I think we could directly use it.

duidae: it has name but name is `人權司法`, not path, 

and I checked neither props do not include `path`
Screen Shot 2023-02-16 at 2 10 45 PM

nor api
Screen Shot 2023-02-16 at 2 12 04 PM

Aylie-Chou: yes, the api returned `name` value storing in database.
`humanrights` is used for fe data mapping.

Aylie-Chou: I found myself misunderstanding 😂
We need path here for the url not wording !

Aylie-Chou (approved):
LGTM 🙇

PR #359, opened 2023-02-17
fix: replace Object.fromentries() to avoid version compatiblity issue

duidae:
twreporter-react升級node到16除了要改circle ci config外應該還要改其他地方, 而且最好需要測試, 而且url.parse那個warning改掉後routing會有問題, 還需要其他處理, 先把Object.fromentries替換成相容的寫法

### DIFF
--- a/packages/core/src/constants/category-set.js
+++ b/packages/core/src/constants/category-set.js
@@ -55,12 +55,12 @@ export const CATEGORY_ID = {
 }
 
 // flip CATEGORY_ID to a {id: path} object
-let flippedCategoryID = {}
+let categoryIDToPath = {}
 Object.entries(CATEGORY_ID).forEach(([key, value]) => {
-  flippedCategoryID[value] = key
+  categoryIDToPath[value] = key
 })
 export const GET_CATEGORY_PATH_FROM_ID = id => {
-  return flippedCategoryID[id]
+  return categoryIDToPath[id]
 }
 
 const subcategoryPath = {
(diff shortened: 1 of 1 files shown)


packages/core/src/constants/category-set.js
```diff
@@ -55,9 +55,10 @@ export const CATEGORY_ID = {
...
 
 // flip CATEGORY_ID to a {id: path} object
-const flippedCategoryID = Object.fromEntries(
-  Object.entries(CATEGORY_ID).map(([key, value]) => [value, key])
-)
+let flippedCategoryID = {}
+Object.entries(CATEGORY_ID).forEach(([key, value]) => {
+  flippedCategoryID[value] = key
+})
 export const GET_CATEGORY_PATH_FROM_ID = id => {
```
Aylie-Chou, line 62: How about using [`findKey`](https://lodash.com/docs/4.17.15#findKey) function provided by lodash here ?
```
export const GET_CATEGORY_PATH_FROM_ID = id => _.findKey(CATEGORY_ID,  (categoryId) => categoryId === id)

duidae: this is a very simple operation, not sure what would be the benefit to introduce a library here?

Aylie-Chou: IMO there's two benefits:
1. we don't need to store `flippedCategoryID` object.
2. current `core` package has `lodash` depency, we would not need to add other library for it.

Aylie-Chou: I think the trade off would be accepted.
Just one more thing here, the naming `flippedCategoryID` could not be easily understand that its key is `path`, which makes L63 a bit confusing.
Could we change the naming to contain some path-to-id meaning for it ?

duidae: renamed, please have a look, thanks.

Aylie-Chou (approved):
LGTM

PR #369, opened 2023-03-10
fix: patch latest & topic on desktop channel

Aylie-Chou:
This patch fix defect [[defect] header 未顯示「最新」、「深度專題」兩個項目](https://app.asana.com/0/1203699264568808/1203964076256284/f)

### DIFF
--- a/packages/universal-header/src/utils/links.js
+++ b/packages/universal-header/src/utils/links.js
@@ -84,11 +84,11 @@ export const checkPathnameParent = (pathname = '', parent = '') => {
   return _.indexOf(_.split(pathname, '/'), parent) === 1
 }
 
-export const getCategoryLink = (
+export function getLink(
   isExternal = defaultIsExternal,
   releaseBranch = defaultReleaseBranch,
   path = ''
-) => {
+) {
   return __getLink(isExternal, releaseBranch, mainBaseURL, path)
 }
 
@@ -185,23 +185,6 @@ export function getSocialMediaLinks() {
       return res
     },
     {}
-  )
-}
-
-export function getChannelLink(
-  isExternal = defaultIsExternal,
-  releaseBranch = defaultReleaseBranch,
-  targetChannelKey = ''
-) {
-  if (!targetChannelKey || !CHANNEL_PATH[targetChannelKey]) {
-    throw new Error('invalid channel key')
-  }
-
-  return __getLink(
-    isExternal,
-    releaseBranch,
-    mainBaseURL,
-    CHANNEL_PATH[targetChannelKey]
   )
 }
 
--- a/packages/universal-header/src/utils/links.js
+++ b/packages/universal-header/src/utils/links.js
@@ -190,8 +190,19 @@ export function getSocialMediaLinks() {
 
 export function getChannelLinks(
   isExternal = defaultIsExternal,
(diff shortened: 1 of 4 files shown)


packages/universal-header/src/utils/links.js
```diff
@@ -190,8 +190,19 @@ export function getSocialMediaLinks() {
...
-  releaseBranch = defaultReleaseBranch
+  releaseBranch = defaultReleaseBranch,
+  targetChannelKey = ''
 ) {
+  if (targetChannelKey) {
+    return __getLink(
+      isExternal,
+      releaseBranch,
+      mainBaseURL,
+      CHANNEL_PATH[targetChannelKey]
```
duidae, line 201: we may need error handling for `CHANNEL_PATH[targetChannelKey]`

duidae (changes requested):

packages/universal-header/src/utils/links.js
Aylie-Chou: error handle for parameter `targetChannelKey` added ([cbda297](https://github.com/twreporter/twreporter-npm-packages/pull/369/commits/cbda2974b2b9f54f9d3019aa150fad9bfef46774))
also seperate `getChannelKey` & `getChannelKeys` function.

packages/universal-header/src/utils/links.js
```diff
@@ -188,6 +188,23 @@ export function getSocialMediaLinks() {
   )
 }
 
+export function getChannelLink(
+  isExternal = defaultIsExternal,
+  releaseBranch = defaultReleaseBranch,
+  targetChannelKey = ''
+) {
+  if (!targetChannelKey || !CHANNEL_PATH[targetChannelKey]) {
+    throw new Error('invalid channel key')
```
duidae, line 197: here we throw the exception but not catching it, seems does not have the effect of error handling, could just copy other js function's handling (ex: find() of array)), return undefined and check the returned value before use

duidae (changes requested):

packages/universal-header/src/utils/links.js
Aylie-Chou: The parameter `targetChannelKey` would basically use values setting in channel constant file, and would not pass any value from user input, thus I think the throwing error is for the development use.
In this scenario, current code would throw error in twreporter-react SSR. I think it supply enough info for developer to understand where went run.

duidae: once we give `getChannelLink` open access to everyone, developers would possibly put anything in the arguments and we may not assume people to know any implicit rule when calling it. We either pull `CHANNEL_PATH` out of this function for clear or properly handle parameters/return value when doing a function call.

duidae (changes requested):

packages/universal-header/src/utils/links.js
Aylie-Chou: updated ([c7856a1](https://github.com/twreporter/twreporter-npm-packages/pull/369/commits/c7856a10811b9053577bcde36aeb73a69c007784))

- add `getLink` to replace `getChannelLink` & `getCategoryLink` function
- add error handle in Channel component

duidae (approved):

PR #374, opened 2023-03-25
Jason/subscription 2 checkbox

duidae:
# Related Issue

[[會員制] 感興趣的議題頁](https://app.asana.com/0/1203699264568808/1203762836596954/f)

# Root Cause

# Solution

### DIFF
--- a/packages/react-components/src/button/utils/theme.js
+++ b/packages/react-components/src/button/utils/theme.js
@@ -349,17 +349,17 @@ export const getActiveTextButtonTheme = theme => {
       }
     case TEXT_BUTTON_THEME.brand:
       return {
-        color: colorBrand.heavy,
+        color: colorBrand.dark,
         hoverColor: colorBrand.dark,
       }
     case TEXT_BUTTON_THEME.dark:
       return {
-        color: colorGrayscale.gray800,
+        color: colorBrand.heavy,
         hoverColor: colorBrand.heavy,
       }
     case TEXT_BUTTON_THEME.light:
       return {
-        color: colorGrayscale.gray600,
+        color: colorGrayscale.gray800,
         hoverColor: colorGrayscale.gray800,
       }
     case TEXT_BUTTON_THEME.normal:
--- a/packages/react-components/src/checkbox/checkbox.stories.js
+++ b/packages/react-components/src/checkbox/checkbox.stories.js
@@ -1,16 +1,14 @@
 import React from 'react'
-import Badge from '.'
-import { colorBrand } from '@twreporter/core/lib/constants/color'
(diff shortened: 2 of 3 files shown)


Aylie-Chou:
Based on the design, the checkbox background color would need to be transparent when value is false.

// current code on storybook
螢幕截圖 2023-04-11 上午11 37 36

// design
螢幕截圖 2023-04-11 上午11 37 55

packages/react-components/src/checkbox/checkbox.stories.js
```diff
@@ -0,0 +1,22 @@
...
+  },
+}
+
+const Template = args => <Checkbox {...args} />
+export const checkbox = Template.bind({})
+checkbox.args = {
+  label: '藝術',
+  value: false,
+  disabled: false,
+  onChange: () => {},
```
Aylie-Chou, line 21: Could we add a story about how `onChange` works (ex. alert checked/uncheck ) ?

packages/react-components/src/button/utils/theme.js
```diff
@@ -349,17 +349,17 @@ export const getActiveTextButtonTheme = theme => {
       }
     case TEXT_BUTTON_THEME.brand:
       return {
-        color: colorBrand.heavy,
+        color: colorBrand.dark,
```
Aylie-Chou, line 352: question: why would we need to change color here ? (the changed color seems not match with design)

packages/react-components/src/checkbox/index.js
```diff
@@ -0,0 +1,56 @@
...
+
+export const Checkbox = ({
+  value = false,
+  label = '',
+  disabled = false,
+  onChange = () => {},
+  ...props
+}) => {
+  const handleChange = e => {
+    onChange && onChange()
```
Aylie-Chou, line 34: may we pass event(`e`) in to onChange function

Aylie-Chou (changes requested):

packages/react-components/src/checkbox/index.js
duidae: added

duidae:
> Based on the design, the checkbox background color would need to be transparent when value is false.
> 
> // current code on storybook 螢幕截圖 2023-04-11 上午11 37 36
> 
> // design 螢幕截圖 2023-04-11 上午11 37 55

fixed, please have a look, thanks.

packages/react-components/src/button/utils/theme.js
Aylie-Chou: found it !

Aylie-Chou (approved):
LGTM

PR #385, opened 2023-05-07
fix: @font-face should use in root file

Aylie-Chou:
# Issue
`Unexpected syntax : @font-face` error when twreporter-react `make dev`
The main reason is that [@font-face](https://www.w3schools.com/cssref/css3_pr_font-face_rule.php) should be imported on root file.

# Dependency
N/A


liruchen32 (approved):
ＬＧＴＭ

PR #397, opened 2023-05-17
feat: mobile member role card

liruchen32:
# Issue
[[會員制] 會員專區的 mobile 會員卡元件](https://app.asana.com/0/1203699264568808/1204568767509308/f)

# Notice

# Dependency

### DIFF
--- a/packages/core/src/constants/member-role.js
+++ b/packages/core/src/constants/member-role.js
@@ -1,0 +1,16 @@
+export const MEMBER_ROLE = {
+  EXPLORER: 'explorer',
+  ACTION_TAKER: 'action_taker',
+  TRAILBLAZER: 'trailblazer',
+}
+
+export const MEMBER_ROLE_ZH_TW = {
+  explorer: '探索者',
+  action_taker: '行動者',
+  trailblazer: '開創者',
+}
+
+export default {
+  MEMBER_ROLE,
+  MEMBER_ROLE_ZH_TW,
+}
--- a/packages/core/src/constants/member-role.js
+++ b/packages/core/src/constants/member-role.js
@@ -1,7 +1,7 @@
 export const MEMBER_ROLE = {
-  EXPLORER: 'explorer',
-  ACTION_TAKER: 'action_taker',
-  TRAILBLAZER: 'trailblazer',
+  explorer: 'explorer',
+  action_taker: 'action_taker',
+  trailblazer: 'trailblazer',
 }
 
 export const MEMBER_ROLE_ZH_TW = {
--- a/packages/react-components/.storybook/preview.js
+++ b/packages/react-components/.storybook/preview.js
@@ -26,7 +26,6 @@ export const parameters = {
   controls: {
     matchers: {
       color: /(background|color)$/i,
-      // 2023/05/17 disable auto convert to datepicker
       // date: /Date$/,
     },
   },
--- a/packages/react-components/.storybook/preview.js
+++ b/packages/react-components/.storybook/preview.js
@@ -26,7 +26,8 @@ export const parameters = {
   controls: {
     matchers: {
       color: /(background|color)$/i,
-      date: /Date$/,
+      // 2023/05/17 disable auto convert to datepicker
+      // date: /Date$/,
     },
   },
   viewport: {
--- a/packages/react-components/.storybook/preview.js
+++ b/packages/react-components/.storybook/preview.js
@@ -26,7 +26,6 @@ export const parameters = {
   controls: {
     matchers: {
       color: /(background|color)$/i,
-      // date: /Date$/,
     },
   },
   viewport: {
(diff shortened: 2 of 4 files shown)


packages/core/src/constants/member-role.js
```diff
@@ -0,0 +1,16 @@
+export const MEMBER_ROLE = {
+  EXPLORER: 'explorer',
```
Aylie-Chou, line 2: Based on the [js style guide](https://www.notion.so/twreporter/JavaScript-Style-Guide-ca71792d06714d54af5e9ffad850bf3f?pvs=4), it would be better to use lowercase for object keys in `MEMBER_RILE`.

packages/core/src/constants/member-role.js
```diff
@@ -0,0 +1,16 @@
...
+  ACTION_TAKER: 'action_taker',
+  TRAILBLAZER: 'trailblazer',
+}
+
+export const MEMBER_ROLE_ZH_TW = {
```
Aylie-Chou, line 7: Question: why would we need to export this const (seems not used in anywhere)

packages/react-components/.storybook/preview.js
```diff
@@ -26,7 +26,8 @@ export const parameters = {
   controls: {
     matchers: {
       color: /(background|color)$/i,
-      date: /Date$/,
+      // 2023/05/17 disable auto convert to datepicker
```
Aylie-Chou, line 29: Please help to delete these comments XD

packages/react-components/src/mobile-member-role-card/index.js
```diff
@@ -0,0 +1,172 @@
...
+        <TitleImg src={titleUrl} />
+        <MarkImg
+          top={markContainerTop}
+          src={markUrl}
+          translateY={markContainerTranslateY}
+        />
+        <DataContainer ref={dataContainerRef}>
+          {role !== MEMBER_ROLE.EXPLORER && (
+            <div>
+              <div style={{ color: colorGrayscale.gray500 }}>
```
Aylie-Chou, line 138: May we manage css style with `styled-component` or `class`

packages/react-components/.storybook/preview.js
liruchen32: OK

packages/core/src/constants/member-role.js
liruchen32: for `twreporter-react`

packages/core/src/constants/member-role.js
liruchen32: fixed

packages/react-components/src/mobile-member-role-card/index.js
liruchen32: fixed

packages/react-components/src/mobile-member-role-card/index.js
```diff
@@ -0,0 +1,176 @@
...
+  bottom: 0;
+`
+const TextContainer = styled.div`
+  color: ${props => props.color};
+  padding-bottom: ${props => props.paddingBottom || 0}px;
+`
+
+function useWindowSize() {
+  const [size, setSize] = useState([0, 0])
+  useLayoutEffect(() => {
```
Aylie-Chou, line 80: Question: why do we choose to use `useLayoutEffect` instead of `useEffect` here ?

liruchen32: > After all DOM mutations have been completed by React, useLayoutEffect executes synchronously. If you need to measure the DOM (for example, to determine the scroll position or other styles for an element) and then modify the DOM or cause a synchronous re-render by changing the state, this can be helpful.

[resource](https://talent500.co/blog/react-useeffect-v-s-uselayouteffect-hook-key-differences/#What_is_useLayoutEffect_hook_in_React)

packages/core/src/constants/member-role.js
Aylie-Chou: Got it. I think it would be okay if we do not use role name from db for this phase, either.

packages/react-components/.storybook/preview.js
Aylie-Chou: also L29 XD

packages/react-components/src/mobile-member-role-card/index.js
Aylie-Chou: `useLayoutEffect` would block browser repaint, which could cause performance problem ([ref](https://react.dev/reference/react/useLayoutEffect))
I found we use the return value of `useLayoutEffect` inside `useEffect`. We may pay the tradeoff but not get benefit.
Could you help to check if there's any other problem if we modify it to `useEffect` ?

packages/react-components/.storybook/preview.js
liruchen32: OK

packages/react-components/src/mobile-member-role-card/index.js
liruchen32: fixed

Aylie-Chou (approved):
LGTM

PR #398, opened 2023-05-24
fix: use flex for auto fill

liruchen32:
# Issue
fix picture height calculate problem

# Notice

# Dependency


Aylie-Chou (approved):
LGTM

PR #399, opened 2023-06-06
feat: get and set user data

liruchen32:
# Issue
[[會員制] 電子報訂閱變更頁](https://app.asana.com/0/1203699264568808/1204692921007274/f)

# Notice

# Dependency


packages/redux/src/reducers/user.js
```diff
@@ -0,0 +1,60 @@
...
+// lodash
+import get from 'lodash/get'
+import merge from 'lodash/merge'
+
+const _ = {
+  get,
+  merge,
+}
+
+const initState = {
```
Aylie-Chou, line 12: just a note: `role` info may also need to be stored in state

Aylie-Chou (changes requested):
Mostly look good !
Wish the new added redux function could have test cases in next PR~~

Aylie-Chou (approved):
LGTM !!
wish comes true so fast 🚀

PR #400, opened 2023-06-07
feat: add TextField component & deprecate TEXT_BUTTON_THEME

Aylie-Chou:
# Issue
[[會員制] 登入註冊：OTP delay 的 UI 調整](https://app.asana.com/0/1203699264568808/1204692921007280/f)

# Notice
- TextButton
  - add `style` props to make it use like other buttons
  - `style` is under one level of `theme` thus the final version of buttons would have 3*3 color sets
  - `TEXT_BUTTON_THEME` & `TEXT_BUTTON_THEME_PROP_TYPES` also deprecated in this patch

# Dependency
- other place using old TextButton method & constants would need migration


packages/react-components/src/input/components/text-field.js
```diff
@@ -0,0 +1,99 @@
...
+import {
+  colorGrayscale,
+  COLOR_SEMANTIC,
+} from '@twreporter/core/lib/constants/color'
+import { P3 } from '../../text/paragraph'
+
+const textAlign = {
+  [TextAlign.LEFT]: 'left',
+  [TextAlign.CENTER]: 'center',
+}
```
liruchen32, line 16: Why do we need this

packages/react-components/src/input/components/text-field.js
```diff
@@ -0,0 +1,99 @@
...
+  message = '',
+  onSubmit = defaultFunc,
+  onChange = defaultFunc,
+  ...props
+}) => {
+  const handleSubmit = e => {
+    e.preventDefault()
+    e.stopPropagation()
+    onSubmit(e.target.input.value)
+  }
```
liruchen32, line 70: Why does the text field need handleSubmit?

packages/react-components/src/input/stories/text-field.stories.js
```diff
@@ -0,0 +1,29 @@
...
+const onSubmit = value => {
+  console.log('submit', value)
+}
+textField.args = {
+  onSubmit,
+  placeholder: 'test',
+  align: TextField.Align.LEFT,
+  state: TextField.State.DEFAULT,
+  message: '系統文字',
+  type: 'text',
```
liruchen32, line 28: type is not defined

packages/react-components/src/input/components/text-field.js
Aylie-Chou: it defines the corresponding style (css value)
would it be better if we wrap them into a object name `style` to recognize ?
```
const style = {
  textAlign: {
    [TextAlign.LEFT]: ...
    ...
  },
  ...
}

packages/react-components/src/input/components/text-field.js
Aylie-Chou: it's a bonus for browser native form.
For example, browser would fire submit event when clicking enter.
Do you consider it a worth feature to include here ?

packages/react-components/src/input/components/text-field.js
liruchen32: I think we already define this at propTypes and enums. We don’t need to do it again

packages/react-components/src/input/stories/text-field.stories.js
liruchen32: 🥹

packages/react-components/src/input/components/text-field.js
Aylie-Chou: fixed ([f346356](https://github.com/twreporter/twreporter-npm-packages/pull/400/commits/f34635625eb9d638bc0ca03768ecc14c9b5944d2))

packages/react-components/src/input/stories/text-field.stories.js
Aylie-Chou: fixed ([f346356](https://github.com/twreporter/twreporter-npm-packages/pull/400/commits/f34635625eb9d638bc0ca03768ecc14c9b5944d2))

liruchen32 (approved):
LGTM

PR #405, opened 2023-06-19
fix: fix action button theme

liruchen32:
# Issue
[20230512 transparent 的 mobile sidemenu 的按鈕看不到（因為是 transparent theme，但應該用 normal 就好）](https://app.asana.com/0/1203983432530614/1204588802643715/f)

# Notice

# Dependency

### DIFF
--- a/packages/universal-header/src/components/action-button-item.js
+++ b/packages/universal-header/src/components/action-button-item.js
@@ -76,7 +76,8 @@ const ActionButtonItem = ({
 }) => {
   const { theme } = useContext(HeaderContext)
   const buttonTheme =
-    direction === DIRECTION_TYPE.row || theme === THEME.photography
+    (direction === DIRECTION_TYPE.row && theme !== THEME.transparent) ||
+    theme === THEME.photography
       ? theme
       : THEME.noraml
   const actionKey = action.key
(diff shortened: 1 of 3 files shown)


packages/universal-header/src/components/action-button-item.js
```diff
@@ -76,7 +76,8 @@ const ActionButtonItem = ({
 }) => {
   const { theme } = useContext(HeaderContext)
   const buttonTheme =
-    direction === DIRECTION_TYPE.row || theme === THEME.photography
+    (direction === DIRECTION_TYPE.row && theme !== THEME.transparent) ||
```
Aylie-Chou, line 79: The `action button` on header would need to use `transparent ` theme may has direction type `row` & theme `transparent`.
螢幕截圖 2023-06-19 下午1 18 41

Aylie-Chou (changes requested):
Current code would also make header action button to use `normal` theme. (which would need to use transparent theme component)

packages/universal-header/src/components/action-button-item.js
liruchen32: fixed

Aylie-Chou (approved):
LGTM

PR #419, opened 2023-08-15
feat: check if is at account page

liruchen32:
# Issue

# Notice

# Dependency

### DIFF
--- a/packages/universal-header/src/components/header.js
+++ b/packages/universal-header/src/components/header.js
@@ -1,4 +1,4 @@
-import React, { useContext, useState } from 'react'
+import React, { useContext, useState, useEffect } from 'react'
 import styled, { css, keyframes } from 'styled-components'
 import CSSTransition from 'react-transition-group/CSSTransition'
 // context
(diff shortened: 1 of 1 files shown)


packages/universal-header/src/components/header.js
```diff
@@ -259,6 +263,16 @@ const Header = () => {
...
+  const needPrevIconAccountRoute = [
+    '/account/donation-history',
+    '/account/email-subscription',
+  ]
+  const isOnNeedPrevIconAcoountPage = _.some(needPrevIconAccountRoute, el =>
+    _.includes(pathname, el)
+  )
+  const showPrevIcon =
+    isOnArticlePage ||
+    (isOnNeedPrevIconAcoountPage && document.body.clientWidth < 768)
```
Aylie-Chou, line 275: Could we give it a var name or add some comment to explain the reason we need to check `clientWidth` here ?

```
// option 1 example
const mobileOnly = document.body.clientWidth < 768
const showPrevIcon = isOnArticlePage || (isOnNeedPrevIconAcoountPage && mobileOnly)

// option 2 example
(isOnNeedPrevIconAcoountPage && document.body.clientWidth < 768) // only show it on mobile

File: packages/universal-header/src/components/header.js [hunk 2]
```diff
@@ -259,6 +263,16 @@ const Header = () => {
     isHamburgerMenuOpen: showHamburger,
   }
   const isOnArticlePage = _.split(pathname, '/')[1] === 'a'
```
Aylie-Chou, line 265: (nice to have)
could we identify if it's on article page with [predifined path name](https://github.com/twreporter/twreporter-npm-packages/blob/master/packages/core/src/constants/entity-path.js) in core package

packages/universal-header/src/components/header.js
liruchen32: fixed

packages/universal-header/src/components/header.js
liruchen32: fixed

Aylie-Chou (approved):
LGTM

PR #427, opened 2023-08-28
chore: upgrade node to v18.17.1 (LTS)

Aylie-Chou:
# Issue
[[工程] upgrade nodejs to v18
](https://app.asana.com/0/1203699264568808/1204298342406882/f)

# Notice
- vercel deployment setting would also need to be updated after release

# Dependency
N/A


liruchen32 (approved):
LGTM

PR #445, opened 2023-10-05
fix: pillbutton layout with icons

Aylie-Chou:
# Issue
style broken when left/right icon added in pillbutton

# Dependency
N/A


liruchen32 (approved):
LGTM

PR #447, opened 2023-10-16
feat: emit onToggleTabExpanded

liruchen32:
# Issue
[[會員制] 文章頁 bottom banner](https://app.asana.com/0/1203699264568808/1205329976610294/f)

# Notice

# Dependency


packages/react-article-components/src/components/table-of-contents/index.js
```diff
@@ -53,6 +54,12 @@ class TableOfContents extends React.PureComponent {
     })
   }
 
+  componentDidUpdate(_prevProps, prevStates) {
```
Aylie-Chou, line 57: Question: why do we name `_prevProps` with prefix `_` & `prevState` not here ?

liruchen32: because `_prevProps` is not used

Aylie-Chou (approved):
LGTM

PR #477, opened 2023-12-21
fix(react-article-components): filter html tag for text render

liruchen32:
# Issue
[一些文章出現html語法](https://app.asana.com/0/0/1206211289316587/f)

# Notice

# Dependency


Aylie-Chou (approved):
LGTM
nice to have: could we add some comment to memo the case to address here

PR #491, opened 2024-01-22
fix: use first category name in `category_set` for bookmark

Aylie-Chou:
# Issue
use first category name in `category_set` as bookmark category

# Notice
`twreporter-react` repo also has utils function to address this issue ([ref](https://github.com/twreporter/twreporter-react/blob/3424cbb32b2588eb4caf5712161dd2a555c04b40/src/utils/bookmark.js#L16))

# Dependency
N/A


liruchen32 (approved):
LGTM

PR #494, opened 2024-01-24
fix: auto check true

liruchen32:
# Issue
check `isBookmark` if is not server side render

# Notice

# Dependency


Aylie-Chou (approved):
ＬＧＴＭ

PR #505, opened 2024-02-20
fix: turn off MY_READING feature toggle

liruchen32:
# Issue
turn off MY_READING feature for releasing

# Notice

# Dependency


Aylie-Chou (approved):
LGTMxd

PR #507, opened 2024-02-20
fix: change ntch roles

howar31:
# Issue
The NTCH collaboration roles incorrect

# Notice
Keep the old role key for fallback compatibility.  Should be remove in the future.

# Dependency


liruchen32 (approved):
LGTM

PR #526, opened 2024-04-15
Feature/post review feature toggle

liruchen32:
# Issue
[[會員制]重要報導回顧 UI](https://app.asana.com/0/1203699264568808/1206721264374708/f)

# Notice

# Dependency


packages/core/src/constants/color.js
```diff
@@ -50,6 +50,7 @@ export const colorGrayscale = Object.freeze({
 })
 
 export const colorOpacity = Object.freeze({
+  white_0: 'rgba(255, 255, 255, 0.0)',
```
Aylie-Chou, line 53: Could we also add `'` wrapping to make it looks consistent

Aylie-Chou (changes requested):
as commnet ~~

packages/core/src/constants/color.js
Aylie-Chou: 😳

Aylie-Chou (approved):
LGTM

PR #527, opened 2024-04-18
fix: set `showColorText` init value `true`

Aylie-Chou:
# Issue
[[spec change] default showColorText=ture
](https://app.asana.com/0/0/1207095807997119/f)

# Dependency
N/A


liruchen32 (approved):
ＬＧＴＭ

PR #546, opened 2024-05-28
fix: add clickable prop for image

liruchen32:
# Issue
Add the `clickable` prop to prevent triggering the `disable-scroll` class.

# Notice

# Dependency


Aylie-Chou (approved):
LGTM 😆

PR #548, opened 2024-05-31
fix(rac): update embedded-code.js. Avoid `executeScript()` twice

nickhsine:
### Issue
slack msg: https://twreporter.slack.com/archives/C09N6TNN7/p1717123213877979

### Notice
After upgrading react to v18, `EmbeddedCode` component becomes abnormal. One case is that `EmbeddedCode` will `executeScript()` twice; therefore, it will generate duplicate contents.

This patch is a workaround to avoid `executeScript()` twice.

### DIFF
--- a/packages/react-article-components/src/components/body/embedded-code.js
+++ b/packages/react-article-components/src/components/body/embedded-code.js
@@ -234,7 +234,8 @@ const WayPointWrapper = props => {
       onEnter={onEnter}
       onLeave={onLeave}
       fireOnRapidScroll={false}
-      topOffset="-150%"
+      topOffset={5}
+      bottomOffset="-100%"
     >
       <div>
         <EmbeddedCode {...props} ref={embedRef} />
--- a/packages/react-article-components/src/components/body/embedded-code.js
+++ b/packages/react-article-components/src/components/body/embedded-code.js
@@ -25,6 +25,7 @@ const _ = {
 const embedNamespace = {
   infogram: 'infogram',
   twreporter: '__twreporterEmbeddedData',
+  storyTellingReporter: '@story-telling-reporter',
 }
 
 export const Block = styled.div`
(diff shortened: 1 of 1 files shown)


packages/react-article-components/src/components/body/embedded-code.js
```diff
@@ -221,7 +234,7 @@ const WayPointWrapper = props => {
       onEnter={onEnter}
       onLeave={onLeave}
       fireOnRapidScroll={false}
-      topOffset={5}
+      topOffset="-150%"
```
Aylie-Chou, line 237: Question: why do we change the `topOffset` value here ?

nickhsine: Oops... This change is not intended.
Thank you for pointing out.

I need to add `bottomOffset="150%"` instead.
The reason to add `bottomOffset` is that I want to execute scripts earlier. 
Therefore, the ScrollableVideo will be loaded before users scroll into the viewport.

nickhsine: https://github.com/twreporter/twreporter-npm-packages/pull/548/commits/0813b6b5f12b22f6ac8831c13a59f678d400e34f fixes the review comment.

Aylie-Chou: Thanks for the detail explanation.
Just want to confirm that do we need to fire all scripts earlier or just `story-telling` scripts ?

nickhsine: For better user experience, I think we should execute all scripts earlier; therefore, users won't see the content shifting. 
But, I am not sure what side effect would be for that change (add `bottomOffset`).
How do you think?

Aylie-Chou (approved):
LGTM 👍

PR #551, opened 2024-06-12
feat: add tracking section for article components

liruchen32:
# Issue
[[會員制]重要報導追蹤-UI/前端: 文章頁](https://app.asana.com/0/1203699264568808/1207085276870883/f)

# Notice

# Dependency

### DIFF
--- a/packages/react-article-components/src/components/body/infobox.js
+++ b/packages/react-article-components/src/components/body/infobox.js
@@ -22,7 +22,7 @@ const _ = {
   get,
 }
 
-const widthCSS = css`
+const articleBodyWidthCSS = css`
   ${mq.mobileOnly`
     width: calc(248/300*100%);
   `}
@@ -40,8 +40,23 @@ const widthCSS = css`
   `}
 `
 
+const trackingSectionWidthCSS = css`
+  ${mq.mobileOnly`
+    width: 256px;
+  `}
+
+  ${mq.tabletOnly`
+    width: 386px;
+  `}
+
+  ${mq.desktopAndAbove`
+    width: 476px;
+  `}
+`
+
 const Title = styled.div`
-  ${widthCSS}
+  ${props =>
+    props.$forTrackingSection ? trackingSectionWidthCSS : articleBodyWidthCSS}
   color: ${colorGrayscale.gray700};
   line-height: 1.9;
   letter-spacing: 0.7px;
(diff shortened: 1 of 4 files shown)


packages/react-article-components/src/components/body/infobox.js
```diff
@@ -165,9 +183,12 @@ export default class Infobox extends PureComponent {
     )
 
     return mutatedContentHtmlString ? (
-      <Container className={className}>
-        {title ? <Title>{title}</Title> : null}
+      <Container className={className} $forTrackingSection={forTrackingSection}>
```
Aylie-Chou, line 186: Good to use `$` for styled-component only props 👍

packages/react-article-components/src/components/body/tracking-section.js
```diff
@@ -0,0 +1,152 @@
...
+          <H4Gray800 text={title} />
+          {contentJSX}
+        </Content>
+      </ContentContainer>
+    </TrackingSectionContainer>
+  )
+}
+
+TrackingSection.propTypes = {
+  data: PropTypes.objectOf({
```
Aylie-Chou, line 145: Could we use `PropTypes.shape` here since the types of values in data are not the same ([ref](https://github.com/facebook/prop-types/blob/1c9c6311c1fb8778bffc7a1ca70c273ee8a8654d/README.md?plain=1#L115))

packages/react-article-components/src/components/body/tracking-section.js
```diff
@@ -0,0 +1,152 @@
...
+    title: PropTypes.string,
+    publishDate: PropTypes.string,
+    content: PropTypes.arrayOf(predefinedPropTypes.elementData),
+  }),
```
Aylie-Chou, line 149: Does `data` a `isRequired` prop ?
If not, could we add a default value for it

packages/react-article-components/src/components/body/tracking-section.js
liruchen32: fixed

packages/react-article-components/src/components/body/tracking-section.js
liruchen32: fixed

Aylie-Chou (approved):
ＬＧＴＭ

PR #559, opened 2024-06-24
Fix/tracking section and bookmark widget

liruchen32:
# Issue
`tracking-section` for article page & `bookmark-widget` render issue while cause article page crashed

# Notice

# Dependency

### DIFF
--- a/packages/react-components/src/bookmark-widget/index.js
+++ b/packages/react-components/src/bookmark-widget/index.js
@@ -1,4 +1,4 @@
-import { useEffect } from 'react'
+import { useEffect, useCallback } from 'react'
 import { useDispatch, useSelector } from 'react-redux'
 import PropTypes from 'prop-types'
 // @twreporter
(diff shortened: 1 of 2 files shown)


Aylie-Chou (approved):
LGTM

packages/react-components/src/bookmark-widget/index.js
```diff
@@ -98,7 +115,7 @@ const BookmarkWidget = ({ articleMeta, renderIcon, toAutoCheck = true }) => {
         articleSlug
       )
     }
-  }, [articleMeta, isAuthed, toAutoCheck, jwt, userID, dispatch])
+  }, [])
```
Aylie-Chou, line 118: Would we miss to check bookmark status when clicking related post if we remove the dependency of `articleMeta` here ?

Aylie-Chou (commented):
Just have a question about the dependency of useEffect
Please have a look at comment ~~

packages/react-components/src/bookmark-widget/index.js
liruchen32: 不會，因為進到相關文章後這個 component 會重新 mount 一次

Aylie-Chou: Okay~~

Aylie-Chou (approved):
LGTM*2

PR #565, opened 2024-07-04
feat(react-article-components): full screen image version2

liruchen32:
# Issue
[[維運] 攝影點圖放大優化](https://app.asana.com/0/1203699264568808/1207085276870887)

# Notice

# Dependency

### DIFF
--- a/packages/react-article-components/src/components/img-with-placeholder/index.js
+++ b/packages/react-article-components/src/components/img-with-placeholder/index.js
@@ -39,19 +39,19 @@ const CrossIconPos = Object.freeze({
 })
 
 // padding: 16px + width: 24px => 40px
-const IconPadding = 40
+const iconPadding = 40
 
 const CrossIconPosCss = position => {
   switch (position) {
     case CrossIconPos.TOP_RIGHT:
       return css`
-        top: -${IconPadding}px;
+        top: -${iconPadding}px;
         right: 0px;
       `
     case CrossIconPos.RIGHT_TOP:
       return css`
         top: 0px;
-        right: -${IconPadding}px;
+        right: -${iconPadding}px;
       `
     case CrossIconPos.INSIDE:
     default:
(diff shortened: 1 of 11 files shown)


packages/react-article-components/src/components/img-with-placeholder/index.js
```diff
@@ -32,11 +32,39 @@ const objectFitConsts = {
...
+      `
+  }
+}
+
 const ImgContainer = styled.div`
   position: relative;
   overflow: hidden;
   width: 100%;
   ${props => props.$heightString}
+  ${props => (props.$clickable ? 'cursor: zoom-in' : '')};
```
Aylie-Chou, line 67: Could we move `;` inside css syntax like `cursor: zoom-in;` to avoid additional `;` when `$clickable` is false

packages/react-article-components/src/components/img-with-placeholder/index.js
```diff
@@ -138,129 +164,142 @@ const CrossIcon = styled.div`
...
+    if (fullScreenImageRef.current) {
+      const img = fullScreenImageRef.current
+      const imgWidth = img.offsetWidth
+      const imgHeight = img.offsetHeight
+      const vw = window.innerWidth
+      const vh = window.innerHeight
+
+      if (imgWidth === vw) {
+        if (vh - imgHeight < 40 * 2) {
+          // padding: 16px + width: 24px => 40px
```
Aylie-Chou, line 229: Could we make `40px` a variable since it has been used more than three times ?

Aylie-Chou (changes requested):
Good to refactor `Img` to functional component !
Please have a look at comments~

packages/react-article-components/src/components/img-with-placeholder/index.js
liruchen32: fixed

packages/react-article-components/src/components/img-with-placeholder/index.js
liruchen32: fixed

PR #576, opened 2024-07-23
Feature/leading embedded article

liruchen32:
# Issue
[[維運] 文章頁新增版型：滿版版型-UI/前端](https://app.asana.com/0/1203699264568808/1207411823835288/f)

# Notice

# Dependency


packages/react-article-components/src/components/leading/embedded.js
```diff
@@ -0,0 +1,218 @@
...
+          case 'embeddedCode':
+          case 'embeddedcode':
+            const { content } = code
+            if (content.length > 0) {
+              const { caption } = content[0]
+              if (caption) {
+                setCaptionText(caption)
+                setEmbeddedCodeJSX(
+                  // caption will be render at this componet, not child component
+                  <EmbeddedCodeComponent data={code} showCaption={!caption} />
```
Aylie-Chou, line 163: IMO, `<EmbeddedCodeComponent ...>` component could have no need to store in state since the default value of `showCaption` is true.
We could assign `showCaption` value based on `cationText`.

packages/react-article-components/src/components/leading/embedded.js
```diff
@@ -0,0 +1,218 @@
...
+            </TopicTitle>
+          ) : null}
+          {subtitle && <SubTitle text={subtitle} />}
+          <Title text={title} />
+        </TitleBlock>
+        {captionText && (
+          <CaptionBlock>
+            <Caption text={captionText} />
+          </CaptionBlock>
+        )}
```
Aylie-Chou, line 208: could we use `null` instead of `false` is cationText is empty here

Aylie-Chou (changes requested):
as comments~~

packages/react-article-components/src/components/leading/embedded.js
liruchen32: fixed

packages/react-article-components/src/components/leading/embedded.js
liruchen32: fixed

Aylie-Chou (approved):
LGTM

PR #577, opened 2024-08-05
fix: check embeddedCode exist and loading issue

liruchen32:
# Issue
The `EmbeddedCode` component was wrapped with `Waypoint`, so if we return the `loading` component, it won't trigger a render.

# Notice

# Dependency

### DIFF
--- a/packages/react-article-components/src/components/body/embedded-code.js
+++ b/packages/react-article-components/src/components/body/embedded-code.js
@@ -168,10 +168,10 @@ class EmbeddedCode extends React.PureComponent {
             dispatchWindowLoadEvent()
           }
           scriptEle.removeEventListener('load', handleLoad)
+          this.props.handleIsLoaded()
         }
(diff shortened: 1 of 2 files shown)


packages/react-article-components/src/components/body/embedded-code.js
```diff
@@ -169,6 +171,7 @@ class EmbeddedCode extends React.PureComponent {
         }
         scriptEle.addEventListener('load', handleLoad)
         scriptsFragment.appendChild(scriptEle)
+        this.props.handleIsLoaded()
```
Aylie-Chou, line 174: Question: why do we call `handleIsLoaded` func here instead of calling it inside `handleLoad` func ?

Aylie-Chou (commented):
Could we add some detail about what this issue is & the root cause in description 🙏

Aylie-Chou (approved):
LGTM

PR #580, opened 2024-08-13
fix: disable click while header is narrow

liruchen32:
# Issue
[[defect] header 隱藏連結仍可被點擊](https://app.asana.com/0/1203699264568808/1207819101456754/f)

# Notice

# Dependency

### DIFF
--- a/packages/universal-header/src/components/header.js
+++ b/packages/universal-header/src/components/header.js
@@ -160,6 +160,7 @@ const TopRow = styled.div`
     transition: opacity ${animation.step3Duration};
     transition-delay: ${(props) =>
       props.$toUseNarrow ? animation.step3Delay : 0};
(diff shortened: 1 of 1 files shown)


packages/universal-header/src/components/header.js
```diff
@@ -160,6 +160,7 @@ const TopRow = styled.div`
     transition: opacity ${animation.step3Duration};
     transition-delay: ${(props) =>
       props.$toUseNarrow ? animation.step3Delay : 0};
+    pointer-events: ${(props) => (props.$toUseNarrow ? 'none' : 'all')};
```
Aylie-Chou, line 163: what a easy resolution to use css tricks !

Aylie-Chou (approved):
LGTM

PR #588, opened 2024-10-24
fix: set loading false when embedded script loaded

Aylie-Chou:
# Issue
[[維運] embeded 滿版新增 place holder 畫面](https://app.asana.com/0/1203699264568808/1208371474585818/f)

# Notice
`EmbeddedCode` component would fire `load` event on `window` after all script loaded.
and `handleIsLoaded` would be called when it start to load

# Dependency
N/A


liruchen32 (approved):
LGTM

PR #590, opened 2024-11-02
fix: dispatch load event for inline script

Aylie-Chou:
# Issue
[[維運] embeded 滿版新增 place holder 畫面](https://app.asana.com/0/1203699264568808/1208371474585818/f)

# Notice
- inline script would not fire `load` event after loaded, thus we need to dispatch `load` event mannually on the end of the inline script text
- also add `handleFinishLoaded` props for `Embedded` component to avoid misunderstanding of `handleLoaded` props

# Dependency
N/A


liruchen32 (approved):
LGTM

PR #598, opened 2024-12-11
fix: use material symbol svg on tab bar

Aylie-Chou:
# Issue
[[維運] 手機 icon 未即時顯示問題](https://www.notion.so/twreporter/icon-33860a6d7a0c4e59973ea716e01cbbbb?pvs=4)

# Dependency
N/A


liruchen32 (approved):
LGTM

PR #604, opened 2025-01-07
fix: font-size in tracking section should be 16px

Aylie-Chou:
# Issue
[[維運]「後續／迴響」字級異常](https://www.notion.so/twreporter/15f8dbdca93280e39588f27d0604c2bd?pvs=4)

# Dependency
N/A


liruchen32 (approved):
LGTM, 跟 class 說掰掰

PR #610, opened 2025-02-26
chore: remove ts config for react-components

Aylie-Chou:
# Issue
- fix ts import cannot find H4 components

# Notice
- the generated `.d.ts` is incorrect

# Dependency
N/A


liruchen32 (approved):
LGTM

PR #612, opened 2025-03-06
fix(core): add ntch-6 key

liruchen32:
# Issue
[[defect]身份別卡片顯示錯誤](https://www.notion.so/twreporter/defect-1a08dbdca93280639025fea4a16f85c0?pvs=4)

# Notice

# Dependency


Aylie-Chou (approved):
ＬＧＴＭ

PR #621, opened 2025-07-08
Feature/add download icon

liruchen32:
# Issue
[[維運] 贊助紀錄新增線下贊助-前端](https://www.notion.so/twreporter/1e78dbdca93280b688dec450183f0b6b?source=copy_link)

# Notice

# Dependency

### DIFF
--- a/packages/react-components/src/empty-state/empty-state-v2.js
+++ b/packages/react-components/src/empty-state/empty-state-v2.js
@@ -9,24 +9,15 @@ import {
 } from '@twreporter/core/lib/constants/release-branch'
 // components
 import { P1, P2 } from '../text/paragraph'
-import { PillButton } from '../button'
 import { Style } from './enums'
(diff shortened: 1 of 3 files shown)


packages/react-components/src/empty-state/empty-state-v2.js
```diff
@@ -0,0 +1,111 @@
...
+  style = Style.DEFAULT,
+  maxWidth = '280px',
+  title = '',
+  guide = null,
+  buttonComponents = [],
+}) => {
+  let imageUrl = ''
+  let imageWidth = ''
+
+  switch (style) {
```
Aylie-Chou, line 64: could we apply `useMemo` or `useCallback` on imageUrl & imageWidth

Aylie-Chou (changes requested):
just one thing to check, plz have a look at comment ~

PR #622, opened 2025-07-16
chore: upgrade node version to v22.17.0

liruchen32:
# Issue
Vercel not support node v18.x 

# Notice

# Dependency


Aylie-Chou (approved):
LGTM

